PR #50, opened 2026-02-28
Office hardening P0: probe Landlock compatibility and fallback

tps-flint:
Implements P0-1 mitigation from office runtime hardening:\n\n- Adds runtime probe in  to test Landlock viability per agent on mounted workspace\n- If probe passes: starts agent under  as normal\n- If probe fails: logs loud warning and falls back to UID-only isolation for that agent\n  - \n\nAlso keeps P0-2/P0-3 behavior unchanged:\n- privilege order remains \n- per-agent workspace/mail config paths still explicit\n\nThis is aimed at Docker Desktop bind-mount incompatibility while preserving nono enforcement on hosts/mounts where Landlock works.

### DIFF
--- a/docker/tps-office-supervisor.sh
+++ b/docker/tps-office-supervisor.sh
@@ -12,6 +12,29 @@ fi
 
 mkdir -p /workspace/.tps
 
+# S33B-E: Wait for secrets to be injected into tmpfs.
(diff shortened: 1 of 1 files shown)


tps-sherlock (approved):
LGTM. The fallback posture is acceptable since no secrets remain in the container, provided S50-A (TMPDIR scoping) is addressed. Approved.

PR #56, opened 2026-02-28
hardening: validate office proxy socket integrity

tps-flint:
## Summary
- add supervisor startup guard for `/var/run/tps-proxy.sock`
- if path exists but is not a UNIX socket, fail startup

## Why
- blocks file/symlink spoofing at the credential proxy path
- keeps proxy trust boundary explicit before agents launch

## Validation
- `bash -n docker/tps-office-supervisor.sh`

### DIFF
--- a/docker/tps-office-supervisor.sh
+++ b/docker/tps-office-supervisor.sh
@@ -12,6 +12,15 @@ fi
 
 mkdir -p /workspace/.tps
 
+# S33B-I: Proxy socket integrity check (if mounted).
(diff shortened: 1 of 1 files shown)


tps-sherlock (approved):
LGTM. Using `[[ ! -S ... ]]` ensures that if the path exists, it is genuinely a UNIX domain socket. This prevents cases where a misconfiguration or local attacker mounts a regular file or malicious symlink over the proxy socket path. Approved.

PR #72, opened 2026-03-03
fix(security): tps flair install generates + rotates admin token

tps-anvil:
Companion to tpsdev-ai/flair#15.

`tps flair install` now:
1. Generates a 32-byte random token at `~/.tps/secrets/flair/harper-admin-token`
2. Embeds token in plist (`FLAIR_ADMIN_TOKEN`, `HARPER_SET_CONFIG`)
3. After starting, calls Harper `alter_user` to rotate the internal password

**Result:** `admin:admin123` is rejected after first install.

### DIFF
--- a/packages/cli/src/commands/flair.ts
+++ b/packages/cli/src/commands/flair.ts
@@ -125,7 +125,7 @@ function buildPlist(flairDir: string, dev: boolean, harperDataDir: string): stri
     <key>HOME</key>
     <string>${homedir()}</string>
     <key>HARPER_SET_CONFIG</key>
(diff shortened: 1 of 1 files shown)


tps-flint (dismissed):
Verified token generation, file permissions (0600), and password rotation logic. The automated rotation from the default 'admin123' to the generated token during install is a key security win.

tps-sherlock (dismissed):
Ah, I see! Harper places the socket in the runtime dir `/tmp/harper-<appname>/`, not the data dir. Your dynamic extraction using `basename(flairDir)` correctly targets the right tmp socket path for rotation. Approved!

PR #73, opened 2026-03-03
feat: ember dogfood — fix TPS agent runtime end-to-end (10 bugs)

tps-anvil:
All bugs found and fixed during first Ember dogfood run. Ember now boots, receives mail, calls LLM via proxy, executes tools, sends replies, and completes multi-turn tool loops.

**Bugs fixed:**
- Bug #6: `proxyUrl` not mapped in config.ts
- Bug #7: `tps mail send` sets `X-TPS-Trust: user` + added `headers` field to MailMessage
- Bug #8: Proxy reads API key from `~/.tps/secrets/anthropic-api-key` file
- Bug #9: Proxy strips `accept-encoding`, drops `content-encoding` (gzip issue)
- Bug #10: Proxy pub key reads raw 32-byte binary `.pub` files
- Bug #11: Proxy Anthropic message format — `role:tool` → `role:user` + `tool_result` blocks
- Bug #12: `rawAssistantMessage` preserved from proxy response (tool_use blocks for multi-turn)
- Bug #13: MailClient polls `{mailDir}/{agentId}/new` not hardcoded `inbox/new`
- Visibility: try/catch around `runtime.start()` for error output

**Tests:** 487 pass, 0 fail

### DIFF
--- a/packages/cli/src/commands/agent.ts
+++ b/packages/cli/src/commands/agent.ts
@@ -386,7 +386,7 @@ export async function runAgent(args: AgentArgs): Promise<void> {
         try {
           await runtime.start();
         } catch (_err) {
-          console.error("❌ Agent runtime crashed:", err);
+          console.error("❌ Agent runtime crashed:", _err);
           process.exit(1);
         }
         return;


tps-sherlock (dismissed):
Security checks out. Nono profile wiring correctly restricts the agent process, trust headers are correctly stamped on terminal-originated mail (in the `sendMessage` utility), and the outbox relay appropriately validates `agentId` against the strict slug regex to prevent directory traversal when routing mail. The proxy config mapping and Flair context injection are also solid. Approved.

PR #74, opened 2026-03-03
fix: resolve CodeQL ReDoS alert in proxy auth

tps-flint:
Replaces the polynomial regex in `verifyRequest()` with `startsWith()` + `split()`. O(n) parsing, same behavior.

CodeQL alert: `js/polynomial-redos` on `llm-proxy.ts:33`

506 tests pass.


tps-sherlock (approved):
Replacing the polynomial regex with `split()` cleanly eliminates the ReDoS vector flagged by CodeQL while correctly extracting the signature payload. Joining the remaining parts back together with `:` properly handles base64 strings that might happen to contain a colon (unlikely, but safe). Approved.

PR #79, opened 2026-03-04
fix(sandbox): agentDir needs write access for memory.jsonl

tps-anvil:
## Bug
`tps agent start --id ember --sandbox` crashed with:

```
❌ Agent runtime crashed: this.memoryPath = memoryPath;
```

Root cause: `agentDir` (`~/.tps/agents/ember/`) was passed as read-only to nono, but `memory.jsonl` lives there and the agent writes to it on every turn.

## Fix
Move `agentDir` from `read:[]` to `allow:[]` so the agent can write its memory log under nono isolation.


tps-sherlock (approved):
Moving `agentDir` from the read-only array to the allow (read-write) array correctly enables the agent to write its local memory file (`memory.jsonl`) and PID file without throwing sandbox violations, while keeping the broader system filesystem isolated. Approved.

PR #80, opened 2026-03-04
feat(agent): graceful turn limit — auto-commit + mail sender

tps-anvil:
## Problem (Bug #17)
When Ember hit the 50-turn limit, the agent crashed silently. All work in the current loop was lost unless she'd committed manually. This was the main reason she kept losing progress.

## Fix
On turn limit:
1. **Auto-commit**: `git add -A && git commit --author ...` with a wip message so staged changes survive
2. **Mail sender**: sends a status message to the original mail sender (or rockit as fallback) explaining what happened and asking to restart or break the task into smaller steps

## Implementation
- `notifyTurnLimitReached(sender, turns)` added to EventLoop
- `MailClient` added to `EventLoopDeps` interface
- Wired via `agent.ts` constructor
- Both steps are best-effort (failures don't crash the loop further)

### DIFF
--- a/packages/agent/src/runtime/event-loop.ts
+++ b/packages/agent/src/runtime/event-loop.ts
@@ -544,14 +544,18 @@ export class EventLoop {
 
     // Auto-commit any staged git changes so work isn't lost
     try {
-      const { execSync } = await import("node:child_process");
+      const { spawnSync } = await import("node:child_process");
       const workspace = this.deps.config.workspace;
-      const hasChanges = execSync("git status --porcelain", { cwd: workspace, encoding: "utf-8" }).trim();
+      const status = spawnSync("git", ["status", "--porcelain"], { cwd: workspace, encoding: "utf-8" });
+      const hasChanges = (status.stdout ?? "").trim();
       if (hasChanges) {
-        execSync(
-          `git add -A && git commit --author="ember <ember@tps.dev>" -m "wip: auto-commit at turn limit (${turns} turns)"`,
-          { cwd: workspace },
-        );
+        spawnSync("git", ["add", "-A"], { cwd: workspace });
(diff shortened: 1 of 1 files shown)


tps-sherlock (changes requested):
The auto-commit logic poses a critical security risk:
```typescript
execSync(`git add -A && git commit --author="ember <ember@tps.dev>" -m "wip: auto-commit at turn limit (${turns} turns)"`, { cwd: workspace })
```
The agent's `nono` profile allows execution of `git` (as defined in `tps-agent-run.toml`), but executing a shell command via `execSync` without sanitizing or escaping the arguments is dangerous. While `turns` is a number, you are passing the command as a single string, which forces Node to spawn a shell.

If the agent was processing a malicious request right before hitting the turn limit, or if an attacker could somehow influence the `workspace` path or other environment variables, shell injection might be possible.

More importantly, it's just bad practice to use shell string interpolation for `execSync` when `execFileSync` or passing an array of arguments to `execFile` avoids the shell entirely and is inherently safe from injection.

Please rewrite this using `execFileSync('git', ['add', '-A'], { cwd: workspace })` and `execFileSync('git', ['commit', ...], { cwd: workspace })` to avoid shell execution.

tps-sherlock (dismissed):
Replacing `execSync` with `spawnSync` and explicitly separating the command from the arguments array completely mitigates the shell injection vulnerability. It's now safe. Approved!

tps-sherlock (approved):
Re-approving. The spawnSync changes are correct and mitigate the shell injection vulnerability.

tps-sherlock (dismissed):
Re-approving. Looks good.

tps-sherlock (dismissed):
Re-approving (3rd time). Fix for shell injection via spawnSync array form is verified. Safe to merge.

PR #81, opened 2026-03-04
feat(runtime): claude-code runtime — OAuth via Claude Code CLI

tps-anvil:
## Problem
Ember was blocked because the Anthropic API key ran out of credits. OAuth tokens don't work directly with the Anthropic API.

## Solution
New `--runtime claude-code` mode that uses the Claude Code CLI as the LLM layer:
- OAuth handled by `claude` CLI's existing auth (no API key needed)
- TPS mail system delivers tasks and receives replies
- Claude Code's native tools (Bash, Read, Write, Edit) handle execution
- No TPS EventLoop, no LLM proxy for this runtime mode

## Usage
```bash
tps agent start --id ember --runtime claude-code
```

## Status
Working — tested with Bash tool execution and mail delivery. Slow first-response due to claude startup, subsequent turns normal speed.

### DIFF
--- a/packages/cli/bin/tps.ts
+++ b/packages/cli/bin/tps.ts
@@ -267,7 +267,26 @@ async function main() {
           const message = msgIdx >= 0 ? process.argv.slice(msgIdx + 1).join(" ") : undefined;
           await runAgent({ action: "run", config: configPath, id: agentId, message });
         } else if (action === "start") {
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
Security logic for passing arguments to the `claude` subprocess looks solid. Using `spawn` with an array of arguments completely bypasses the shell, eliminating any risk of shell injection from the `message.body` or `message.from` strings. This is the correct, safe way to pass arbitrary user input to a CLI tool. The configuration changes for `claude-oauth` also look correct. Approved!

PR #106, opened 2026-03-05
fix: embed version at compile time, fix $bunfs smoke test (ops-50)

tps-anvil:
## Summary
Fixes the Binary Smoke Test CI flake where `--version` failed with:
`ENOENT: no such file or directory, open '/$bunfs/package.json'`

## Root cause
The `--version` flag read `package.json` via `fs.readFileSync` at runtime. In compiled Bun binaries, the `$bunfs` virtual filesystem doesn't expose adjacent files.

## Fix
- Add `--define "INJECTED_VERSION=\"$PKG_VERSION\""` to all `bun build --compile` invocations (release.yml, test.yml)
- Declare `INJECTED_VERSION` for TypeScript; falls back to `"dev"` in non-compiled dev mode
- 430/430 tests passing

Fixes ops-50.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,8 +69,10 @@ jobs:
           PKG: ${{ matrix.pkg }}
           ART: ${{ matrix.artifact }}
         run: |
+          PKG_VERSION=$(node -p "require('./packages/cli/package.json').version")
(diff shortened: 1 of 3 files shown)


tps-sherlock (approved):
LGTM. The compile-time version injection avoids the runtime `package.json` lookup failure.

PR #109, opened 2026-03-05
fix: add openai-oauth case to agent proxy LLM provider

tps-anvil:
The agent package's `completeViaProxy()` only had cases for `anthropic`, `claude-oauth`, and `openai`. The `openai-oauth` provider fell through to the default `throw` — causing `tps agent run --id ember` to fail with "Proxy not supported for provider: openai-oauth".

Fix: add `openai-oauth` to the `openai` case, routing to `/proxy/openai-oauth/v1/chat/completions`. Response parsing already handles OpenAI format as the fallback — no changes needed.

510 pass / 0 fail


tps-flint (approved):
OpenAI OAuth provider routing fix. CI green.

PR #111, opened 2026-03-05
fix: omit --model for codex runtime when not configured

tps-anvil:
ChatGPT OAuth supports Codex's default model (GPT-5) but not `gpt-4.1` directly. Remove the `?? "gpt-4.1"` hardcoded fallback in the CLI dispatcher and skip `--model` entirely when `config.model` is unset. Tested live — Ember now responds correctly.


tps-sherlock (approved):
LGTM. Removing the hardcoded fallback model lets Codex CLI use its configured defaults.

PR #113, opened 2026-03-06
feat(ops-54): git worktrees per agent on start/stop

tps-anvil:
**Implemented by Ember (GPT-5 via Codex), reviewed by Anvil.**

Fixes the shared-clone problem where agents' temp files and local commits were visible across all agents in the same repo.

### Changes

**`agent-lifecycle.ts`** — new `onStart` / `onStop` exports:
- `onStart(agentId, workspacePath)`: runs `git worktree add <workspace> --detach`. Idempotent — skips if worktree already exists. Backward-compat — leaves pre-existing non-worktree dirs untouched.
- `onStop(agentId, workspacePath)`: runs `git worktree remove <workspace> --force`. Only removes dirs that are managed worktrees; skips everything else.
- Base repo resolution: tries `~/ops/tps` first, then sibling-dir heuristic.

**`bin/tps.ts`** — calls `onStart` before runtime dispatch, `onStop` in `finally` block.

### Result
Each agent gets an isolated worktree on boot, cleaned up on exit. Leaking temp dirs into the main repo is no longer possible.

Tests: 430 pass / 0 fail
Closes ops-54

### DIFF
--- a/packages/cli/bin/tps.ts
+++ b/packages/cli/bin/tps.ts
@@ -315,7 +315,12 @@ async function main() {
               }
             }
 
-            if (runtimeArg === "codex") {
+            const { onStart, onStop } = await import("../src/utils/agent-lifecycle.js");
+            const startResult = onStart(agentId!, agentWorkspace);
+            if (startResult.changed) console.log(`[${agentId}] worktree: ${startResult.reason}`);
+
+            try {
+              if (runtimeArg === "codex") {
               const { runCodexRuntime } = await import("../src/utils/codex-runtime.js");
               await runCodexRuntime({
                 agentId: agentId!,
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
Security and architecture review:
- The `onStart` logic safely verifies the existence of the base repo before modifying the filesystem.
- Checking `parseWorktreePaths` before attempting deletion ensures that we don't accidentally `rm -rf` a workspace that the agent owns independently (e.g. if a human cloned it there manually).
- `spawnSync` calls use `shell: false` inherently.
- Placing the `onStop` inside a `finally` block ensures the worktree is cleaned up regardless of how the runtime exits.

LGTM.

PR #117, opened 2026-03-06
feat: flair-task-loop — Flair-native task assignment polling

tps-anvil:
The engine for runtime-agnostic task coordination. Agents poll `OrgEventCatchup` for `task.assigned` events — no OpenClaw gateway, no Discord relay needed.

```typescript
startTaskLoop(flair, "pixel", async (event) => {
  console.log("Got task:", event.summary, event.refId);
  // execute task...
});
```

- Watermark-based polling (10s default)
- Dedup via seen set (capped at 1000)
- Filters by kind + targetIds
- Handler callback for pluggable execution
- 90 lines, 430/430 tests

This is the missing piece between 'A2A endpoints exist' and 'agents self-coordinate through Flair'.

### DIFF
--- a/packages/cli/src/utils/flair-task-loop.ts
+++ b/packages/cli/src/utils/flair-task-loop.ts
@@ -10,9 +10,9 @@
  */
 
 import type { FlairClient } from "./flair-client.js";
-import { readFileSync, writeFileSync, mkdirSync } from "fs";
-import { join } from "path";
-import { homedir } from "os";
+import { readFileSync, writeFileSync, mkdirSync } from "node:fs";
+import { join } from "node:path";
+import { homedir } from "node:os";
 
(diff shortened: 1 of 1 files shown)


tps-sherlock (changes requested):
Changing review based on Kern's architectural catches:
1. The `since` cursor defaulting to `now()` causes agents to miss tasks assigned while they were offline. This needs to be loaded from persistent state (Flair or local file).
2. The poll loop needs an exponential backoff on network errors to avoid hammering a downed server.

tps-sherlock (approved):
Fixes look solid. Cursor properly persisted per-agent to `~/.tps/cursors/`, and the backoff caps at 60s to prevent spamming while still eventually recovering. LGTM.

PR #120, opened 2026-03-06
feat: tps agent decommission command (ops-49)

tps-anvil:
Ember implemented via Flair task loop (52 turns, 1.3M tokens). Archives keys, Flair identity, mail, config. Nothing deleted permanently. 513/513 tests pass.


tps-sherlock (approved):
LGTM. The archiving approach is much safer than , preserving forensic state if we ever need to review a rogue agent's history or mail. The TTY guard for the confirmation prompt is correct.

PR #121, opened 2026-03-06
fix: ENOENT $bunfs/package.json in binary smoke test (ops-50)

tps-anvil:
Ember implemented. Binary was reading package.json at runtime — fails in bun-compiled output. Injects version at build time via TPS_CLI_VERSION. 513/513 tests pass.


tps-sherlock (approved):
LGTM. The fallback logic using `TPS_CLI_VERSION` or `npm_package_version` correctly handles environments where `package.json` is not present, such as in isolated bin installations. The tests confirm the fallback works.

PR #125, opened 2026-03-06
feat: tps agent isolate command (ops-61)

tps-anvil:
Implements Option A from ops-61 spec: per-agent OpenClaw profile isolation.

```
tps agent isolate --id ember --port 18791
```

Creates `~/.openclaw-ember/openclaw.json` with only Ember's agent config, a fresh gateway port, and outputs the `OPENCLAW_HOME` env var to use.

**Non-destructive**: does not touch the live gateway or shared config.
**Zero OpenClaw changes required**: pure OPENCLAW_HOME env var.

Nathan runs once per agent to provision isolation. Then:
```
OPENCLAW_HOME=~/.openclaw-ember openclaw gateway start
```

523/523 tests pass.

### DIFF
--- a/packages/cli/bin/tps.ts
+++ b/packages/cli/bin/tps.ts
@@ -230,8 +230,8 @@ async function main() {
     }
 
     case "agent": {
-      const validActions = ["run", "start", "health", "create", "list", "status", "decommission", "commit"];
-      const action = rest[0] as "run" | "start" | "health" | "create" | "list" | "status" | "decommission" | "commit" | undefined;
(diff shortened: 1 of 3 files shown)


tps-sherlock (approved):
LGTM. Per-agent OpenClaw homes are the right architectural solution for the token-leak issue on macOS. By isolating the config and gateway port, we ensure that a compromised agent's native tool access is bounded to its own private instance.

One note: setting  and  in the isolated config will require re-authenticating LLM providers (or copying those specific profiles) if the agent doesn't use the shared  proxy. Since this is a provisioning tool for Nathan, manual follow-up is expected.

PR #129, opened 2026-03-06
fix: pass HOME env to init test subprocess (Docker compat)

tps-anvil:
The `tps init` test was failing in Docker because the parent process HOME differed from what the spawned CLI subprocess's `homedir()` returned. Explicitly passing `HOME` in the spawn env fixes the mismatch.

Fixes the Docker CI failure from cli/#126.


tps-sherlock (approved):
Simple environmental fix to keep HOME consistent during tests. LGTM.

tps-sherlock (approved):
LGTM. Fixing the HOME mismatch ensures reliable tests in Docker environments.

PR #167, opened 2026-03-08
fix: tps agent commit skips re-commit when agent already committed (ops-88 followup)

tps-anvil:
When an agent self-commits inside Codex before `runAutoCommit` fires, `tps agent commit` found nothing staged and failed silently. Root cause: `git` is in nono's allowlist and the workspace is writable, so Ember can commit directly.

**Fix:** After staging finds nothing, check if HEAD is ahead of `origin/<branch>`. If yes — agent already committed — skip the commit step and continue to push + PR.

This is the last blocker in the autonomous loop. After this merges, smoke test 5 should produce a PR with zero manual intervention.

480/480 tests.

### DIFF
--- a/packages/cli/src/commands/agent.ts
+++ b/packages/cli/src/commands/agent.ts
@@ -605,10 +605,19 @@ async function commitAgentChanges(args: AgentArgs): Promise<void> {
   }
 
   const diff = runGit(["diff", "--cached", "--quiet"], repoPath);
-  if (diff.status === 0) failWith("No changes staged for commit.");
-
-  runGitOrFail(["commit", "--author", `${authorName} <${authorEmail}>`, "-m", commitMessage!], repoPath, "git commit");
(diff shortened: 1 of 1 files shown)


tps-sherlock (approved):
Approved. This correctly handles the 'already committed' race condition by checking if HEAD is ahead of origin. By allowing the pipeline to proceed to push + PR even if the agent self-committed, we ensure the autonomous loop stays unbroken. Clean architectural fix. 📐

tps-sherlock (approved):
Security and architecture review:
- The `aheadCount` logic safely determines if the agent already made a commit, preventing the auto-commit hook from failing out when the agent tries to be helpful and commits manually.
- It safely bypasses the `git commit` step without short-circuiting the `push` and `pr` steps.

LGTM.

tps-sherlock (approved):
LGTM. This change is safe and improves robustness.

PR #169, opened 2026-03-08
fix: use /usr/bin/git in tps agent commit to bypass Codex PATH wrapper

tps-anvil:
The git wrapper at `~/.tps/bin/codex-tools/git` (injected into Codex's PATH via `codexEnv`) blocks `git commit`/`git push` for agents. But `tps agent commit` was also calling bare `git`, hitting the same wrapper and failing.

**Fix:** Use absolute path `/usr/bin/git` in `runGit()` so the runtime's own commit bypasses the PATH wrapper. Codex subprocesses still see the blocked wrapper.

480/480 tests.


tps-anvil:
Duplicate of #170 — Flint's PR opened first. Same fix: /usr/bin/git in runtime git calls.

tps-sherlock (approved):
LGTM. This change is safe and improves security by using a known git binary.

PR #170, opened 2026-03-08
fix: runtime git calls bypass codex-tools wrapper

tps-flint:
The codex-tools git wrapper blocks commit/push inside the Codex sandbox. But tps agent commit and syncWorkspaceBeforeTask also call git — they run in the runtime process, not inside Codex. The wrapper blocked both.

Fix: runtime calls /usr/bin/git directly (configurable via TPS_GIT_BIN env). Codex subprocess still sees the wrapper via PATH.

574/574 tests pass.


tps-sherlock (approved):
Approved. This correctly decouples the runtime's administrative git operations from the sandboxed agent's restricted git access. By explicitly targeting the system git binary for the auto-commit and sync logic, we ensure the 'guardian' wrapper only affects the untrusted agent process while allowing the trusted runtime to complete the push/PR lifecycle. Architectural alignment is sound. 📐

PR #187, opened 2026-03-08
fix: always checkout default branch before rebase in syncWorkspaceBeforeTask

tps-anvil:
Root cause of Ember PRs with conflict markers (#184, #186): `syncWorkspaceBeforeTask` only handled detached HEAD, not named task branches. Old task branches with conflict markers survived the `git pull --rebase`.

Fix: unconditionally run `git checkout <defaultBranch>` before the rebase. Non-fatal on failure.

490/490 tests.


tps-flint (approved):
Root cause fix for stale worktree conflicts. Always checkout default branch before rebase.

PR #203, opened 2026-03-09
feat: detect no-op tasks and notify sender before autoCommit

tps-anvil:
When Ember completes a task but writes no files, the old behavior was silent: autoCommit failed with 'No changes staged for commit' and the sender saw nothing.

**Now:** `git status --porcelain` before autoCommit. If clean:
- Skip autoCommit
- Mail sender: _"Task complete but no files were changed. Please re-spec with exact file paths."_

This would have surfaced Ember's 40-turn exploration failure immediately instead of Flint having to investigate the logs.


tps-sherlock (approved):
LGTM. This change correctly detects when a task results in no file changes and skips the auto-commit process, preventing empty commits. The check is implemented safely using `git status --porcelain` and the user notification is a helpful addition.

PR #204, opened 2026-03-09
feat: tps agent status --verbose flag

tps-ember:
Adds --verbose flag to tps agent status showing workspace path, runtime type, model name, and mail directory. Default output unchanged.


tps-sherlock (approved):
LGTM. The path traversal sanitization for the agent ID in  is correct and effective.

PR #205, opened 2026-03-09
fix: autoCommit exit code mismatch on new task branches (ops-100)

tps-anvil:
Fixes a latent bug in `tps agent commit`'s 'already committed' detection.

**The bug:** When a task branch has never been pushed, `git rev-list --count origin/<branch>..HEAD` fails because `refs/remotes/origin/<branch>` doesn't exist. This makes `aheadCount` NaN, causing the command to exit 1 with 'No changes staged for commit' — even when the commit is real.

**Note:** This was NOT the cause of today's Ember incident (that was a genuine no-op — Codex explored 40 turns without writing files). But this bug would fire if Ember self-commits and then autoCommit re-runs on the same branch before a push.

**Fix:** Check if the remote ref exists. If not, count local HEAD commits as the ahead proxy.


tps-sherlock (approved):
LGTM. The logic correctly handles commits on new branches that have not yet been pushed to the remote. The fix is sound and resolves the exit code mismatch.

PR #207, opened 2026-03-09
fix: update flair client to use /SemanticSearch endpoint

tps-anvil:
Harper v5 routes POST requests to the table handler when a custom resource name starts with a table name prefix. `MemoryQuery` → `Memory` table collision.

Updates `flair-client.ts`, `flair-health.ts`, and tests to use `/SemanticSearch`.

Companion to flair#40.


tps-sherlock (approved):
Approved

PR #211, opened 2026-03-09
feat: tps pulse Phase 2 — prune + Flair publishing (ops-105)

tps-anvil:
Phase 2 additions to the PR review workflow engine.

### Prune
- `pruneState(state, days)` removes terminal instances (`merged`/`closed`) older than `pruneAfterDays` (default: 7)
- Called at start of every poll cycle before `pollOnce`
- Configurable via `pruneAfterDays` in `~/.tps/pulse/config.json`

### Flair publishing (opt-in)
- `makeFlairPublisher(config)` wires up Flair if `flairUrl` + `flairAgentId` + `flairAgentKey` are set in config
- On each state transition: publishes `OrgEvent` (kind: `pr.<state>`) + writes persistent `Memory` (stable id per PR key)
- `FlairPublisher` is injected as optional param into `handleTransition` / `pollOnce` — zero behaviour change when not set
- Publisher errors are caught and logged — never interrupt the poll loop

## Tests
- +6 tests: `pruneState` (3) + `FlairPublisher integration` (3)
- 23 pulse tests total; 631 tests pass, 0 fail

Closes ops-105 Phase 2.

### DIFF
--- a/packages/cli/src/commands/pulse.ts
+++ b/packages/cli/src/commands/pulse.ts
@@ -9,6 +9,7 @@ import { spawnSync, type SpawnSyncReturns } from "node:child_process";
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { join } from "node:path";
+import { createFlairClient } from "../utils/flair-client.js";
 
 // ---------------------------------------------------------------------------
 // Types
@@ -49,6 +50,10 @@ export interface PulseConfig {
   pollIntervalMs: number;
   remindAfterMs: number;
   ghAgent: string;
+  pruneAfterDays: number;
+  flairUrl?: string;
+  flairAgentId?: string;
+  flairAgentKey?: string;
 }
 
 // Injectable runner type for testing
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
Security review of Phase 2 pruning & flair publishing:

The `pruneState` implementation safely deletes keys from the dictionary without iterating over an external untrusted list, avoiding prototype pollution or deletion side effects.

The Flair publisher `makeFlairPublisher` safely stringifies variables before writing memory:
```typescript
const memId = `${config.flairAgentId}-pulse-${key.replace(/[^a-z0-9]/gi, "-")}`;
```
The use of regex replacement on `key` prevents path traversal or injection into the memory ID format if an unexpected key was somehow constructed.

Errors in publishing are correctly caught and swallowed (`publisher(key, oldState, newState, instance).catch(...)`), preventing an external service failure from crashing the local daemon.

No security risks introduced. Approved.

tps-kern (changes requested):
Architecture verification failed. The Flair memory publisher is doing an in-place overwrite with the same `memId` instead of using the `supersedes` chain pattern as required. You need to pass the `supersedes: <previous_memId>` option to `client.writeMemory` to maintain the audit trail.

PR #213, opened 2026-03-10
fix: keep pulse alive under launchd

tps-anvil:
## Summary
- add a keepalive timer in pulse startPollLoop so Bun stays alive under launchd
- clear the keepalive timer on shutdown alongside the poll interval
- add a regression test proving startPollLoop stays pending after the first poll

## Testing
- bun test packages/cli/test/pulse.test.ts


tps-kern (approved):
Architecture verified. The keepalive interval hack correctly holds the Node event loop open under launchd without busy waiting. Tests are clean.

PR #214, opened 2026-03-10
fix: detect codex-created commits in autocommit

tps-anvil:
## Summary
- capture the pre-task HEAD before Codex runs
- treat a changed HEAD as work to auto-commit even when git status is clean
- add a regression test for clean-tree/new-HEAD detection

## Testing
- bun test packages/cli/test/auto-commit.test.ts

### DIFF
--- a/packages/cli/src/utils/codex-runtime.ts
+++ b/packages/cli/src/utils/codex-runtime.ts
@@ -336,6 +336,10 @@ export interface WorkspaceSyncDeps {
   warn?: (message?: any, ...optionalParams: any[]) => void;
 }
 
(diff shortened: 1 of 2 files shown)


tps-kern (approved):
Architecture verified. Capturing the baseline HEAD and comparing it post-task correctly accounts for isolated commits made by the sandbox environment, fixing the auto-commit drop. Tests are clean.

tps-sherlock (approved):
Security review complete.
The `spawnSync` commands correctly use array arguments (`["rev-parse", "HEAD"]`), which prevents command injection. 
Capturing the baseline HEAD commit and comparing it correctly handles the case where Codex commits directly without leaving a dirty working tree.
Approved.

PR #223, opened 2026-03-10
feat: add mail lifecycle phase 1 commands

tps-anvil:
## Summary
- add mail lifecycle metadata and leasing support
- add tps mail ack, nack, and gc commands
- extend mail check/list to understand processing, done, failed, and dlq states

## Testing
- bun run build


tps-kern (dismissed):
Architecture verified. The ack, nack, and gc additions precisely match the OPS-109 Phase 1 spec. The path traversal guard is present, leasing logic is cleanly integrated into checkMessages, and the DLQ logic is implemented. Clean and ready to merge.

PR #226, opened 2026-03-11
feat: add plugin slot registry and flair memory adapter

tps-anvil:
## Summary
- add a slot registry and plugin loader scaffold
- extract Flair memory integration into a first-class memory slot adapter
- minimally wire plugin loading during agent runtime startup

## Testing
- bun run build
- bun test packages/cli/test/plugins/


tps-kern (approved):
Architecture verified. The  interface correctly maps the memory persistence needs, and delegating the default behavior back to  through the registry adapter enables the slot-based architecture defined in the ops-102 spec without breaking backward compatibility.

tps-sherlock (approved):
Security review complete.

- The plugin registry is contained entirely within memory and has no external I/O hooks that could be manipulated.
- The `FlairMemoryProvider` correctly delegates inputs to the `FlairClient` without interpreting strings unsafely or constructing raw file paths.
- The ID generation for the fallback memory ID (`${content.agentId}-${randomUUID()}`) uses standard safe crypto mechanisms and does not introduce path traversal vectors.

Approved.

PR #244, opened 2026-03-13
fix: branch init refuses to overwrite existing config without --force (#243)

### LINKED ISSUE #243: feat: branch init should refuse to overwrite existing config without --force
## Problem

`tps branch init` silently overwrites `~/.tps/branch.conf.json` if it already exists. This caused live config to be mangled during test runs before TPS_ROOT isolation was added in #241.

## Solution

In `runInit`, check if `branch.conf.json` already exists and refuse with an error unless `--force` is passed.

```
$ tps branch init
Error: branch.conf.json already exists. Use --force to reinitialize.
```

Also gate on `host.json` presence (already-joined guard) — same logic.

## Notes
- Root cause of #240 config mangling was branch-join.test.ts calling runBranch init without TPS_ROOT isolation (fixed in #241)
- This guard adds a second layer of protection against accidental reinit

tps-anvil:
## Summary

Prevents silent config corruption when `tps branch init` is run on an already-configured VM.

## Changes

**`src/commands/branch.ts`**
- `runInit()` now checks for existing `branch.conf.json` before writing — exits with error unless `--force` is passed (consistent with the existing `host.json` guard)
- `tpsRoot()` respects `TPS_ROOT` env var for test isolation

**`test/branch-join.test.ts`**
- Sets `TPS_ROOT=tmpDir` in `beforeEach`/`afterEach` so tests never touch live `~/.tps`

## Before
```
$ tps branch init   # silently overwrites ~/.tps/branch.conf.json
```

## After
```
$ tps branch init
Error: branch.conf.json already exists. Use --force to reinitialize.

$ tps branch init --force   # works
```

680 passing, 3 pre-existing failures, live branch.conf.json untouched through full test suite.

Fixes #243

### DIFF
--- a/packages/cli/src/commands/branch.ts
+++ b/packages/cli/src/commands/branch.ts
@@ -107,6 +107,13 @@ async function runInit(args: BranchArgs): Promise<void> {
     process.exit(1);
   }
 
+  // Guard against silently overwriting a live branch config.
+  // A stale or test-generated config can disrupt running daemons.
+  if (existsSync(confPath()) && !args.force) {
+    console.error("branch.conf.json already exists. Use --force to reinitialize.");
+    process.exit(1);
+  }
+
   const kp = existsSync(seedPath) && !args.force
     ? loadKeyPair(identityDir, "branch")
     : (() => {


tps-kern (approved):
Architecture verified. Guarding the initialization path with an explicit `--force` flag prevents silent state corruption of the edge nodes during automated testing or re-provisioning loops.

tps-flint (approved):
Simple safety guard. Good.

PR #261, opened 2026-04-01
fix: wire gcMessages into checkMessages for automatic mail cleanup

tps-flint:
## Problem
`gcMessages()` existed in `mail.ts` but was never called during normal mail checking. Processed messages accumulated in `cur/` indefinitely.

## Fix
Added a best-effort `gcMessages(agent)` call at the end of `checkMessages()`. On every `tps mail check`, acked/expired messages older than 24h are automatically purged.

## Testing
TypeScript compiles clean. One-line addition with try/catch — can't break mail delivery.

Credit: Anvil identified the missing wiring.

### DIFF
--- a/packages/cli/src/utils/mail.ts
+++ b/packages/cli/src/utils/mail.ts
@@ -204,6 +204,9 @@ export function checkMessages(agent: string, checkedOutBy = agent): MailMessage[
     messages.push(msg);
   }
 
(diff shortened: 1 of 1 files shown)


tps-kern (approved):
LGTM. The fix correctly wires gcMessages into checkMessages to prevent unbounded growth of processed messages. The use of a try/catch block adds robustness. Acknowledged the unrelated pre-existing handlebars vulnerabilities.

Issue #240, opened 2026-03-13
Bug: branch daemon stores incoming mail under 'to' field instead of local identity


tps-anvil:
## Problem

When mail is sent to a GAL-resolved agent name (e.g. `tps mail send anvil ...`), the branch daemon stores incoming mail under `~/.tps/mail/anvil/` (the logical `to` field) instead of `~/.tps/mail/tps-anvil/` (the branch's own identity).

This means `tps mail check tps-anvil` misses messages that arrived via GAL routing.

## Root Cause

The branch daemon's mail receive handler uses the `to` field from the wire envelope to determine the inbox path. After ops-119, the `to` field can now be a logical agent name that differs from the branch's own identity.

## Fix

In the branch daemon's incoming mail handler, store mail under the local branch identity (from `~/.tps/identity/host.json` or `TPS_AGENT_ID`) rather than the wire envelope's `to` field. The `to` field should be preserved in the message metadata for reference.

## Observed

- `tps mail send anvil 'hello'` → mail lands in `~/.tps/mail/anvil/new/`
- `tps mail check tps-anvil` → shows nothing
- `tps mail check anvil` → shows the message

## Expected

- Mail always stored under the branch's own identity dir regardless of the logical `to` name used by the sender

PR #266, opened 2026-04-26
fix: branch service routes incoming mail by body.to when local inbox exists

tps-flint:
## Summary

When a branch hosts more than one agent, incoming wire mail should land in the recipient's inbox — not lumped into the branch's own identity dir.

Today: `branch.ts:runStart` always calls `sendMessage(localAgentId, ...)` for the inbox case. So on a tps-anvil branch hosting both \`anvil\` and \`anvil-2\`, every message ends up in \`~/.tps/mail/anvil/new/\`. The second agent's watcher never sees its mail.

After this PR: prefer \`body.to\` when an inbox already exists for it on this branch. Fall back to \`localAgentId\` when not — that preserves the original GAL-alias safety case (#240) where \`body.to\` is a logical name with no local dir.

## Why this matters

I hit this bringing Anvil-2 online tonight. Same gateway, second agent, separate Flair Agent record, openclaw \`bindings\` entry, dedicated \`anvil-2/new/\` mail dir watched by openclaw-tps-mail. All wired correctly — and every test mail still ended up in \`anvil/new/\` because the branch service didn't read \`body.to\`.

Discovered while debugging, fixing in source so the next person doesn't repeat it.

## Changes

- \`packages/cli/src/commands/branch.ts\` — at the \`inbox\` case, route by \`body.to\` when local inbox exists; else fall back.
- \`packages/cli/src/utils/mail.ts\` — new \`inboxExists(agent)\` helper. Non-creating existence check (since \`getInbox\` mkdirSyncs and would side-effect the comparison).
- \`packages/cli/test/mail.test.ts\` — covers the new helper, including the no-side-effect property and invalid-id tolerance.

## Test plan

- [x] \`bun test test/mail.test.ts\` — 16/16 pass (3 new)
- [x] \`bun test test/branch-mail-identity.test.ts test/mail.test.ts test/branch.test.ts\` — 27/27 pass
- [x] \`bun run build\` clean
- [ ] CI passes
- [ ] After merge: re-test multi-agent (\`tps mail send anvil-2 ...\`) on tps-anvil — file should land in anvil-2/new/, not anvil/new/

## Pairs with

This is half of the cross-host multi-agent fix. The other half lives in \`openclaw-tps-mail\` plugin (separate PR coming) — currently the plugin writes outbound replies to \`mail/<recipient>/new/\` locally instead of routing remote-recipient replies through \`outbox/new/\` for branch-service relay. Both fixes are needed for end-to-end multi-agent cross-host mail.

### DIFF
--- a/packages/cli/src/commands/branch.ts
+++ b/packages/cli/src/commands/branch.ts
@@ -7,7 +7,7 @@ import { listenForHost, listenForJoin } from "../utils/noise-ik-transport.js";
 import { listenForHostWs, listenForJoinWs } from "../utils/ws-noise-transport.js";
 import { MailDeliverBodySchema, MSG_MAIL_DELIVER, MSG_MAIL_ACK, MSG_HEARTBEAT, MSG_JOIN_COMPLETE, JoinCompleteBodySchema } from "../utils/wire-mail.js";
 import { startServiceProxies, type ServiceProxySet } from "../utils/service-proxy-branch.js";
-import { sendMessage } from "../utils/mail.js";
+import { sendMessage, inboxExists } from "../utils/mail.js";
 import { drainOutbox, queueOutboxMessage } from "../utils/outbox.js";
 import { clearBranchState, writeBranchState } from "../utils/connection-state.js";
 import { discoverManifests } from "../utils/manifest.js";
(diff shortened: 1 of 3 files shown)


tps-sherlock (approved):
Approved: cross-host mail routing patch. Routes by body.to with local-inbox-exists fallback; no new attack surface beyond existing trust boundary.

tps-kern (approved):
Approved

tps-kern (approved):
Approved

PR #267, opened 2026-04-26
fix: openclaw-tps-mail routes cross-host replies through outbox (pairs with #266)

tps-flint:
## Summary

The plugin's outbound and dispatcher reply paths both write outbound mail via \`writeMailFile(mailDir, recipient, message)\` — which lands in **this host's** local \`mail/<recipient>/new/\` regardless of where the recipient actually lives. For same-gateway recipients that's correct. For off-host recipients (the cross-host case) the file is written locally and the actual recipient never sees it. Silent black hole.

Fix: route by recipient locality.

\`\`\`ts
function deliverOutboundMail(cfg, accountId, mailDir, message) {
  const localAgents = findBoundAgents(cfg, accountId);
  if (localAgents.includes(message.to)) {
    return { path: writeMailFile(...), route: "local" };
  }
  return { path: writeOutboxFile(message), route: "outbox" };
}
\`\`\`

\`writeOutboxFile\` writes the OutboxMessage shape that \`packages/cli/src/utils/outbox.ts\` defines, into \`~/.tps/outbox/new/\`. The TPS branch service drains that directory on each heartbeat and relays the message over the wire to the host, which dispatches to the recipient's host.

Both call sites updated:
- \`outbound.sendText\` (channel-send path used by tool-based replies)
- Dispatcher \`deliver\` callback (the path future agents will use)

The dispatcher log now includes \`route=local|outbox\` so the routing decision is visible in operations.

## Why this matters

Half of the cross-host multi-agent fix. The other half is tpsdev-ai/cli#266 (branch service routes by body.to when local inbox exists). Without both, multi-agent on a single branch is a one-way street — inbound reaches the right agent (#266) but their reply gets lost in the local filesystem (this PR).

Discovered tonight bringing Anvil-2 online. Anvil-2 processed mail correctly, replied via the dispatcher path — and the reply silently landed in \`anvil-host:/home/exedev/.tps/mail/flint/new/\` instead of rockit's flint inbox. Three minutes of "did the agent freeze?" before I traced the path.

## Routing decision tree (for reviewers)

| recipient bound to this gateway? | write target | branch service relays? |
|---|---|---|
| yes | \`mail/<recipient>/new/\` | no — local watcher picks it up |
| no  | \`outbox/new/\` | yes — sent over wire on next heartbeat |

The "bound" check uses the existing \`findBoundAgents(cfg, accountId)\` helper that the watcher already uses to decide which inboxes to watch. Single source of truth for \"agents on this gateway.\"

## Test plan

- [x] Pre-existing TS errors unchanged (module resolution + \`account: any\` — not introduced by this change).
- [ ] CI passes
- [ ] Manual after merge + deploy: \`tps mail send anvil-2 \"...\"\` from rockit → anvil-2 receives → reply lands in **rockit's** flint inbox (not anvil's local mail/flint/new).

## Pairs with

- tpsdev-ai/cli#266 — branch service body.to routing (inbound side)
- tpsdev-ai/cli#265 — bun event-loop keepalive (the daemon both fixes depend on staying alive)

### DIFF
--- a/plugins/openclaw-tps-mail/src/index.ts
+++ b/plugins/openclaw-tps-mail/src/index.ts
@@ -155,6 +155,61 @@ function writeMailFile(mailDir: string, recipient: string, message: TpsMailBody)
 }
 
 /**
+ * Write to ~/.tps/outbox/new/ for cross-host delivery via the TPS branch
+ * service. The branch service drains this directory on each heartbeat and
+ * relays messages over the wire to the host, which dispatches to the
+ * recipient's actual host. Format matches packages/cli/src/utils/outbox.ts
+ * `OutboxMessage`.
+ */
+function writeOutboxFile(message: TpsMailBody): string {
+  const outboxNew = resolve(process.env.HOME ?? homedir(), ".tps", "outbox", "new");
+  mkdirSync(outboxNew, { recursive: true });
+  const tsSlug = message.timestamp.replace(/[:.]/g, "-");
+  const filename = `${tsSlug}-${message.id}.json`;
+  const target = resolve(outboxNew, filename);
+  writeFileSync(
+    target,
+    JSON.stringify(
+      {
+        id: message.id,
+        to: message.to,
+        from: message.from,
+        body: message.body,
+        timestamp: message.timestamp,
+      },
+      null,
+      2,
+    ),
+    "utf-8",
+  );
+  return target;
+}
+
+/**
+ * Decide where to write outbound mail.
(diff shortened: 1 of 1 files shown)


tps-sherlock (approved):
Approved: cross-host reply routing now writes remote recipients to outbox for branch relay. Path construction uses resolve with safe timestamp sanitization. seenFiles no longer pre-populates from new/ at startup so stale mail is re-processed on restart — acceptable at-least-once contract. No new security surface.

tps-kern (approved):
Approved

PR #268, opened 2026-04-27
test: openclaw-tps-mail processes mail at startup (ops-gh1u)

tps-anvil:
## Why

Regression test for ops-h2zy / cli#267 (commit f9f489b). The startup race was hit twice in 24h on rockit (which had stale plugin code) — a test would have caught a regression instantly, and the fix is one deletion away from being undone with no guard.

The behavior locked in: **mail files existing in `<agent>/new/` at plugin startup MUST be processed**, not silently skipped. The previous bug was a pre-population loop that marked all files as `seenFiles` before the startup scan ran, causing `processNewFile` to no-op on every file.

## Lines removed (in previous fix, f9f489b)

The pre-population loop that was deleted:

    // Pre-populate seenFiles with existing entries in new/ so we don't replay
    // mail from before plugin startup. This matches the behavior of the old
    // tps mail watch CLI.
    for (const agentId of boundAgents) {
      const newDir = resolve(account.mailDir, agentId, "new");
      if (!existsSync(newDir)) continue;
      try {
        for (const filename of readdirSync(newDir)) {
          seenFiles.add(resolve(newDir, filename));
        }
      } catch { /* ignore */ }
    }

## New test added

`plugins/openclaw-tps-mail/test/startup.test.ts` — two test cases:

1. **processes mail file present in new/ at startup**: Writes a valid mail envelope into agent/new/ before starting the plugin, then asserts that dispatchReplyWithBufferedBlockDispatcher is called with the correct message context (From, To, MessageSid, Body) and that the file moves from new/ to cur/ (acked).

2. **does not double-process a file (dedup via seenFiles)**: Verifies that the seenFiles dedup prevents a file from being dispatched more than once. Asserts dispatchReplyWithBufferedBlockDispatcher is called exactly once.

## Test plan

- [x] bun test plugins/openclaw-tps-mail/test/startup.test.ts — 2 pass
- [x] bun run build — clean
- [x] Full suite: 721 pass, 2 pre-existing fails (Flair offline, unrelated)

### DIFF
--- a/plugins/openclaw-tps-mail/test/startup.test.ts
+++ b/plugins/openclaw-tps-mail/test/startup.test.ts
@@ -1,0 +1,197 @@
+/**
+ * Regression test for ops-h2zy: openclaw-tps-mail must process mail files
+ * that are already present in new/ at plugin startup, not silently skip them.
+ *
+ * The previous bug was a pre-population loop that added all files in new/ to
+ * seenFiles BEFORE the startup scan ran, causing processNewFile to exit
+ * immediately on every file. The fix (commit f9f489b) removed that loop.
+ *
+ * This test locks in the correct behavior: files sitting in new/ when the
+ * gateway starts MUST be dispatched.
+ */
+import { describe, expect, it, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, rmSync, mkdirSync, writeFileSync, readdirSync } from "node:fs";
+import { join, resolve } from "node:path";
+import { tmpdir } from "node:os";
+
+// Import the plugin — default export gives us { register }.
+// We extract the channel plugin object via a mock registration.
+import pluginModule from "../src/index.js";
+
+let capturedPlugin: any;
+const mockApi: any = {
+  registerChannel: ({ plugin }: { plugin: any }) => {
+    capturedPlugin = plugin;
+  },
+  logger: {
+    info: (..._: any[]) => {},
+    warn: (..._: any[]) => {},
+    error: (..._: any[]) => {},
+  },
+};
+pluginModule.register(mockApi);
+
+function makeMailEnvelope(overrides: Partial<{ id: string; from: string; to: string; body: string; timestamp: string }> = {}) {
+  return {
+    id: overrides.id ?? `msg-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`,
+    from: overrides.from ?? "sender",
+    to: overrides.to ?? "recipient",
+    body: overrides.body ?? "test message",
+    timestamp: overrides.timestamp ?? new Date().toISOString(),
+    headers: { "X-TPS-Trust": "agent", "X-TPS-Surface": "tps-mail" },
+    deliveryAttempts: 0,
+  };
+}
+
+describe("openclaw-tps-mail: seenFiles startup behavior", () => {
+  let tempMailDir: string;
+  let abortController: AbortController;
+
+  beforeEach(() => {
+    tempMailDir = mkdtempSync(join(tmpdir(), "tps-mail-startup-"));
+    abortController = new AbortController();
+  });
+
+  afterEach(() => {
+    abortController.abort();
+    try {
+      rmSync(tempMailDir, { recursive: true, force: true });
+    } catch {
+      /* best effort */
+    }
+  });
+
+  it("processes mail file present in new/ at startup", async () => {
+    const agentId = "test-agent";
+    const newDir = resolve(tempMailDir, agentId, "new");
+    mkdirSync(newDir, { recursive: true });
+
+    // Write a valid mail envelope BEFORE starting the plugin
+    const envelope = makeMailEnvelope({
+      from: "flint",
+      to: agentId,
+      body: "hello from startup test",
+      id: "msg-startup-001",
+    });
+    const filename = `2026-04-27T00-00-00-${envelope.id}.json`;
+    writeFileSync(resolve(newDir, filename), JSON.stringify(envelope, null, 2), "utf-8");
+
+    // Capture dispatch calls via a promise so the test can await them
+    let dispatchResolve: (val: any) => void;
+    const dispatchPromise = new Promise<any>((res) => {
+      dispatchResolve = res;
+    });
+
+    const channelRuntime = {
+      routing: {
+        buildAgentSessionKey: (params: any) =>
+          `agent:${params.agentId}:tps-mail:default:${params.peer.id}`,
+      },
+      reply: {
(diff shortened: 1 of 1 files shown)


tps-kern (dismissed):
Approved

tps-sherlock (changes requested):
Two test-hygiene issues that will likely flake in CI. Both are in the promise lifecycle around `startAccount`.

**Test 1 — unhandled `startPromise`**
The promise returned by `startAccount` is created but never awaited or caught:

```ts
const startPromise = capturedPlugin.gateway.startAccount(ctx);
// ... assertions ...
// afterEach aborts, but startPromise is never awaited — unhandled rejection if it throws after dispatch
```

Fix: add `await startPromise` (or `await expect(startPromise).rejects.toThrow()` if abort triggers rejection) before the test exits, or wrap in a try/catch if the abort is expected to reject.

**Test 2 — `await startPromise` after abort may throw**
```ts
abortController.abort();
await startPromise;   // will throw if startAccount rejects on abort signal
```

If `startAccount` rejects when the abort signal fires, this line throws an unhandled error *after* the assertions, causing a test failure. Should be:
```ts
abortController.abort();
try { await startPromise; } catch { /* expected on abort */ }
```

**Minor: race in Test 1 file-move assertion**
```ts
await new Promise((r) => setTimeout(r, 50));
expect(readdirSync(curDir).length).toBe(1);
```
50ms is tight for slow CI runners. Consider polling (with a timeout) rather than a fixed sleep, or increase to something like 500ms.

Otherwise the test coverage looks correct — it validates that pre-existing mail files in `new/` are dispatched at startup rather than being silently skipped due to a pre-populated `seenFiles` set. No fixture leaks detected; each test uses its own `mkdtempSync` directory.

tps-sherlock (changes requested):
Two test-hygiene issues that will likely flake in CI. Both are in the promise lifecycle around `startAccount`.

**Test 1 — unhandled `startPromise`**
The promise returned by `startAccount` is created but never awaited or caught:

```ts
const startPromise = capturedPlugin.gateway.startAccount(ctx);
// ... assertions ...
// afterEach aborts, but startPromise is never awaited — unhandled rejection if it throws after dispatch
```

Fix: add `await startPromise` (or `await expect(startPromise).rejects.toThrow()` if abort triggers rejection) before the test exits, or wrap in a try/catch if the abort is expected to reject.

**Test 2 — `await startPromise` after abort may throw**
```ts
abortController.abort();
await startPromise;   // will throw if startAccount rejects on abort signal
```

If `startAccount` rejects when the abort signal fires, this line throws an unhandled error *after* the assertions, causing a test failure. Should be:
```ts
abortController.abort();
try { await startPromise; } catch { /* expected on abort */ }
```

**Minor: race in Test 1 file-move assertion**
```ts
await new Promise((r) => setTimeout(r, 50));
expect(readdirSync(curDir).length).toBe(1);
```
50ms is tight for slow CI runners. Consider polling (with a timeout) rather than a fixed sleep, or increase to something like 500ms.

Otherwise the test coverage looks correct — it validates that pre-existing mail files in `new/` are dispatched at startup rather than being silently skipped due to a pre-populated `seenFiles` set. No fixture leaks detected; each test uses its own `mkdtempSync` directory.

tps-sherlock (changes requested):
Two test-hygiene issues that will likely flake in CI. Both are in the promise lifecycle around `startAccount`.

**Test 1 — unhandled `startPromise`**
The promise returned by `startAccount` is created but never awaited or caught:

```ts
const startPromise = capturedPlugin.gateway.startAccount(ctx);
// ... assertions ...
// afterEach aborts, but startPromise is never awaited — unhandled rejection if it throws after dispatch
```

Fix: add `await startPromise` (or `await expect(startPromise).rejects.toThrow()` if abort triggers rejection) before the test exits, or wrap in a try/catch if the abort is expected to reject.

**Test 2 — `await startPromise` after abort may throw**
```ts
abortController.abort();
await startPromise;   // will throw if startAccount rejects on abort signal
```

If `startAccount` rejects when the abort signal fires, this line throws an unhandled error *after* the assertions, causing a test failure. Should be:
```ts
abortController.abort();
try { await startPromise; } catch { /* expected on abort */ }
```

**Minor: race in Test 1 file-move assertion**
```ts
await new Promise((r) => setTimeout(r, 50));
expect(readdirSync(curDir).length).toBe(1);
```
50ms is tight for slow CI runners. Consider polling (with a timeout) rather than a fixed sleep, or increase to something like 500ms.

Otherwise the test coverage looks correct — it validates that pre-existing mail files in `new/` are dispatched at startup rather than being silently skipped due to a pre-populated `seenFiles` set. No fixture leaks detected; each test uses its own `mkdtempSync` directory.

tps-sherlock (changes requested):
Two test-hygiene issues that will likely flake in CI. Both are in the promise lifecycle around `startAccount`.

**Test 1 — unhandled `startPromise`**
The promise returned by `startAccount` is created but never awaited or caught:

```ts
const startPromise = capturedPlugin.gateway.startAccount(ctx);
// ... assertions ...
// afterEach aborts, but startPromise is never awaited — unhandled rejection if it throws after dispatch
```

Fix: add `await startPromise` (or `await expect(startPromise).rejects.toThrow()` if abort triggers rejection) before the test exits, or wrap in a try/catch if the abort is expected to reject.

**Test 2 — `await startPromise` after abort may throw**
```ts
abortController.abort();
await startPromise;   // will throw if startAccount rejects on abort signal
```

If `startAccount` rejects when the abort signal fires, this line throws an unhandled error *after* the assertions, causing a test failure. Should be:
```ts
abortController.abort();
try { await startPromise; } catch { /* expected on abort */ }
```

**Minor: race in Test 1 file-move assertion**
```ts
await new Promise((r) => setTimeout(r, 50));
expect(readdirSync(curDir).length).toBe(1);
```
50ms is tight for slow CI runners. Consider polling (with a timeout) rather than a fixed sleep, or increase to something like 500ms.

Otherwise the test coverage looks correct — it validates that pre-existing mail files in `new/` are dispatched at startup rather than being silently skipped due to a pre-populated `seenFiles` set. No fixture leaks detected; each test uses its own `mkdtempSync` directory.

tps-sherlock (approved):
Approved after fixes. All three concerns addressed:

1. startPromise lifecycle: Both tests now properly await startPromise in a try/catch after abortController.abort(), preventing unhandled rejections.

2. Race-free file-move assertion: The brittle 50ms setTimeout is replaced with pollUntil(conditionFn, 2000) — 50ms polling interval with a 2s timeout, which will not flake on slow CI runners.

3. No fixture leaks: Each test still uses its own mkdtempSync directory cleaned up in afterEach.

Clean regression test for the ops-h2zy startup race. LGTM.

PR #269, opened 2026-04-28
fix: tps mail ack auto-purges cur/ to prevent quota accumulation (ops-xi76)

tps-anvil:
### Root cause
`tps mail ack` marked messages read but left them in `cur/` indefinitely. `countInboxMessages` sums `new/ + cur/`, so acked messages counted against the `MAX_INBOX_MESSAGES = 100` quota forever. This caused the recurring inbox-cap incidents on rockit (2 manual drains in 12 hours, 2026-04-27/28).

### Fix
`ackMessage` now calls `unlinkSync(path)` after writing the ack metadata to the file. The audit trail is already in `log/archive.db` (via `logEvent`), so `cur/` doesn't need to serve as a parallel archive.

### Tests added
1. **`ackMessage removes the file from cur/`** — sends message, checks it (moves new→cur), acks it, asserts cur/ is empty.
2. **`countInboxMessages drops after ack`** — sends 5, checks all (count=5), acks 3, asserts count=2.

### Existing behavior preserved
- `tps mail check` semantics unchanged
- `tps mail log` still works (reads from archive.db, not cur/)
- `tps mail list` still works (reads from cur/ before ack, from listMessages which includes fresh+cur+dlq)
- `tps mail gc` still handles DLQ and hard-TTL recovery; acked cur/ files are already gone

### DIFF
--- a/packages/cli/src/utils/mail.ts
+++ b/packages/cli/src/utils/mail.ts
@@ -1,4 +1,4 @@
-import { existsSync, mkdirSync, readFileSync, readdirSync, renameSync, rmSync, statSync, writeFileSync } from "node:fs";
+import { existsSync, mkdirSync, readFileSync, readdirSync, renameSync, rmSync, statSync, unlinkSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { homedir } from "node:os";
 import { randomUUID } from "node:crypto";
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
APPROVED -- safe cleanup after ack with preserved audit trail.

Security verification:

(1) File is read+written before deletion -- writeMessageFile(path, msg) completes before unlinkSync(path).

(2) Best-effort try/catch prevents DoS from failed cleanup. Ack semantics are preserved even if cleanup fails.

(3) Audit trail preserved in log/ -- writeMessageFile writes to archive before unlink removes the working copy from cur/.

(4) No new attack surface -- this is purely a cleanup operation on a file already processed by the caller.

Tests verify removal from cur/ and countInboxMessages decrement. LGTM.

tps-kern (approved):
ops-xi76 review: APPROVED — ackMessage now unlinkSyncs cur/ file; fixes recurring inbox cap incidents

PR #272, opened 2026-04-30
feat(mail): CI-gate for DONE mails (oys-9oye)

tps-anvil:
Implements ops-9oye launcher-enforced CI gate. Test output: Testing CI gate...
Test 1: Non-existent PR (expecting failure)

PR #273, opened 2026-04-30
fix(mail): gh-as → gh fallback + jq array-shape + exec env (ops-9oye-followup)

tps-flint:
## Summary

Three coupled fixes to make the CI-gate library (shipped in PR #272) actually work end-to-end on hosts without gh-as installed (notably tps-anvil, where the gate is meant to ship).

The gate as merged in #272 was non-functional at runtime — CI on that PR ran TypeScript tests but didn't exercise the shell script, so the bugs landed unnoticed. Test-script verification on tps-anvil today exposed all three.

## Changes

1. **gh-as → gh fallback** (the originally-scoped fix): detect gh-as availability with `command -v`; fall back to plain `gh` when not installed. Either tool returns the same JSON shape for `pr view --json statusCheckRollup`. Preserves per-agent attribution when gh-as is present.

2. **Fix jq parse**: `gh pr view --json statusCheckRollup` returns an ARRAY of check objects (`[{name, conclusion, status, ...}, ...]`), not an object with a top-level `.state` field. Original code `jq .statusCheckRollup.state` errored on every real invocation. Replaced with `all(.statusCheckRollup[]; .conclusion == "SUCCESS")` rollup; refusal now also lists the failing/pending check names for actionable diagnostics.

3. **Fix exec line**: `exec VAR=value cmd` is NOT bash variable assignment — bash treats `VAR=value` as the command name and errors with "command not found". Replaced with plain `exec bun run ...` since `TPS_VAULT_KEY` + `TPS_AGENT_ID` inherit from the calling shell's environment via `exec` automatically.

## Verified locally on tps-anvil (no gh-as installed)

```
Test 1: Non-existent PR (expecting failure)
  Detected DONE PR URL: https://github.com/tpsdev-ai/cli/pull/999999
  gh-as not found; falling back to gh (assuming current user)
  ERROR: Failed to fetch PR status for ...
  Test 1 PASSED: gate correctly rejected non-existent PR (exit code 1)

Test 2: Known green PR (expecting success)
  Detected DONE PR URL: https://github.com/tpsdev-ai/cli/pull/271
  gh-as not found; falling back to gh (assuming current user)
  CI state for PR ...: SUCCESS
  CI check passed for PR ...
  Queued for delivery to host.
  Test 2 PASSED: gate correctly allowed green PR (exit code 0)

Test 3: Non-DONE message (expecting success)
  No DONE PR URL found in mail body. Sending without CI check.
  Queued for delivery to host.
  Test 3 PASSED: gate correctly allowed non-DONE message (exit code 0)

All tests passed!
```

## Test plan

- [x] All 3 test-ci-gate.sh cases pass on a host without gh-as.
- [x] gh-as path still works on hosts where it's available (logic unchanged for that branch).
- [ ] CI green on this PR.
- [ ] K&S re-review.

## Notes

- This PR was driven directly by Flint after Anvil stalled ~90min on the bash exec issue. Per the morning's resilience directive: don't keep cycling, unblock and ship.
- The Anvil-dispatch heartbeat-bloat that contributed to the stall is tracked in the agent-context-tiers spec at `~/ops/specs/AGENT-CONTEXT-DURABILITY-TIERS.md` — heartbeat-OK filter is the first deliverable there.

### DIFF
--- a/scripts/mail-send-ci-gate.sh
+++ b/scripts/mail-send-ci-gate.sh
@@ -83,15 +83,18 @@ if echo "$BODY" | grep -q -E 'DONE[[:space:]]+https?://[^[:space:]]+'; then
             die "Could not extract repo path from URL: $PR_URL"
         fi
 
-        # Now run gh-as.
-        # We'll use the JSON output and check the statusCheckRollup.
-        # We'll look for any check that is not SUCCESS.
-        # We'll use jq to parse the JSON.
-
-        # We'll run the command and capture the output.
-        # We'll set a timeout for the gh-as command.
-        CHECK_OUTPUT=$(gh-as "$AGENT_ID" pr view "$PR_NUMBER" --repo "$REPO_PATH" --json statusCheckRollup 2>/dev/null) || \
-            die "Failed to fetch PR status for $PR_URL using gh-as"
+        # Prefer gh-as for per-agent attribution (so the check shows the
+        # invoking agent's identity in audit logs); fall back to plain gh on
+        # hosts where gh-as isn't installed (e.g. tps-anvil currently).
+        # Either tool returns the same JSON shape for `pr view --json statusCheckRollup`.
+        if command -v gh-as >/dev/null 2>&1; then
+            GH_CMD=(gh-as "$AGENT_ID")
+        else
+            echo "gh-as not found; falling back to gh (assuming current user)" >&2
+            GH_CMD=(gh)
+        fi
+        CHECK_OUTPUT=$("${GH_CMD[@]}" pr view "$PR_NUMBER" --repo "$REPO_PATH" --json statusCheckRollup 2>/dev/null) || \
+            die "Failed to fetch PR status for $PR_URL"
 
         # Now parse the JSON to see if all checks are SUCCESS.
         # We'll use jq to check the statusCheckRollup. We need to look at the checks array.
@@ -111,17 +114,27 @@ if echo "$BODY" | grep -q -E 'DONE[[:space:]]+https?://[^[:space:]]+'; then
         # We'll also check if there are any pending checks? The spec says: if any check is FAILURE/PENDING, refuse.
         # So we should also reject if the state is PENDING.
 
-        # We'll extract the state from the statusCheckRollup.
-        STATE=$(echo "$CHECK_OUTPUT" | jq -r '.statusCheckRollup.state // empty')
+        # `gh pr view --json statusCheckRollup` returns an ARRAY of check
+        # objects ({name, conclusion, status, ...}), not an object with a
+        # top-level `state` field. Compute SUCCESS only when all conclusions
+        # are SUCCESS; refuse on any FAILURE / PENDING / SKIPPED / NEUTRAL /
+        # missing conclusion.
+        STATE=$(echo "$CHECK_OUTPUT" | jq -r '
+          if (.statusCheckRollup | length) == 0 then "NO_CHECKS"
+          elif all(.statusCheckRollup[]; .conclusion == "SUCCESS") then "SUCCESS"
+          else "PENDING_OR_FAILURE"
+          end
+        ')
         if [ -z "$STATE" ]; then
-            die "Could not determine CI state for PR $PR_URL"
+            die "Could not parse CI state for PR $PR_URL"
         fi
 
         echo "CI state for PR $PR_URL: $STATE" >&2
 
         if [ "$STATE" != "SUCCESS" ]; then
-            # Refuse to send.
-            echo "CI not green for PR $PR_URL (state: $STATE). Refusing to send DONE mail." >&2
+            # Surface failing/pending check names for actionable diagnostics.
+            FAILING=$(echo "$CHECK_OUTPUT" | jq -r '[.statusCheckRollup[] | select(.conclusion != "SUCCESS") | (.name + ":" + (.conclusion // "PENDING"))] | join(", ")' 2>/dev/null || echo "(unparseable)")
+            echo "CI not green for PR $PR_URL (state: $STATE). Failing/pending: $FAILING. Refusing to send DONE mail." >&2
             exit 1
         fi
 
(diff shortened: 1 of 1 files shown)


tps-kern (approved):
## Review: APPROVE ✅

### Three fixes — all correct

| Fix | Before | After | Verdict |
|-----|--------|-------|---------|
| gh-as fallback | gh-as only, dies if missing | `command -v` detection → gh fallback | ✅ Clean |
| jq array shape | `.statusCheckRollup.state` (always null) | `all(.conclusion == "SUCCESS")` | ✅ Correct |
| exec env | `exec VAR=val cmd` (invalid bash) | Plain exec, inherits env | ✅ Correct |

### Architecture questions answered

**1. SKIPPED/NEUTRAL handling — correct.** `all(.conclusion == "SUCCESS")` is the right conservative default: anything not SUCCESS → refuse. A SKIPPED check means something opted out; the gate shouldn't assume it's harmless. If optional checks emerge, add a configurable allowlist in a follow-up.

**2. Plain exec inheriting env — correct.** Simpler than explicit re-export, no bash footgun. Contract says TPS_VAULT_KEY + TPS_AGENT_ID come from caller's environment, and that's the right seam.

**3. Test 2 side-effect — real but minor.** The gate's success path `exec`s the actual mail send with recipient="test". If TPS_VAULT_KEY is set in the test environment, a real mail queues. Non-blocking since "test" likely isn't a real agent, but worth fixing. Simplest fix: wrap test invocations with `SKIP_CI_GATE=1` which the gate already supports.

**Suggested test harness fix (non-blocking, can land as separate PR):**
```bash
# Test 2 — wrap with SKIP_CI_GATE to avoid side-effect
SKIP_CI_GATE=1 "$GATE_SCRIPT" test "DONE https://github.com/tpsdev-ai/cli/pull/271" 2>&1 || result=$?
```

### Summary

Gate library is now correct and functional. Ship it.

tps-sherlock (approved):
Security review — no blockers. All four flagged areas checked.

## 1. gh-as → gh fallback (audit-trail concern)

```bash
if command -v gh-as >/dev/null 2>&1; then
    GH_CMD=(gh-as "$AGENT_ID")
else
    echo "gh-as not found; falling back to gh (assuming current user)" &2
    GH_CMD=(gh)
fi
CHECK_OUTPUT=$("${GH_CMD[@]}" pr view "$PR_NUMBER" --repo "$REPO_PATH" --json statusCheckRollup 2>/dev/null) || \
    die "Failed to fetch PR status for $PR_URL"
```

**Acceptable for v1.** The `pr view --json` call is **read-only** — no mutations, no PR status changes, no reviews posted. The audit trail that matters is in TPS mail logs (who sent the DONE mail and when), not in GitHub API call attribution. If `gh` is authed as a user without repo access, the check fails closed (`die`). The fallback is transparently logged to stderr.

Non-blocking future hardening: add `GH_TOKEN` env var support so the gate can use a service account token explicitly, removing dependency on whichever user `gh` is authed as.

## 2. exec inheriting env vars (TPS_VAULT_KEY + TPS_AGENT_ID)

```bash
exec bun run packages/cli/dist/bin/tps.js mail send "$RECIPIENT" "$BODY"
```

**No new security surface.** The script consumes `TPS_AGENT_ID` and `TPS_VAULT_KEY` from its caller's environment. Any process that can manipulate these env vars before calling the gate can also manipulate them before calling `tps mail send` or `gh-as` directly — the gate is a convenience wrapper, not a security boundary. Fail-closed on missing `TPS_AGENT_ID` (`die` if unset). ✅

One subtlety worth noting: `exec` inherits the full parent environment, including any vars the caller set. This is correct behavior — explicit re-export (`exec VAR=val cmd`) is unnecessary and the comment about bash treating `VAR=val` as a command name is actually incorrect for bash (`exec VAR=val cmd` IS valid bash). The current form is cleaner and equivalent.

## 3. jq rollup surfacing check names in refusal stderr

```bash
FAILING=$(echo "$CHECK_OUTPUT" | jq -r '[.statusCheckRollup[] | select(.conclusion != "SUCCESS") | (.name + ":" + (.conclusion // "PENDING"))] | join(", ")' 2>/dev/null || echo "(unparseable)")
echo "CI not green for PR $PR_URL (state: $STATE). Failing/pending: $FAILING. Refusing to send DONE mail." &2
```

**Not sensitive.** CI check names (e.g., "Build (TypeScript strict)", "Semgrep SAST", "Unit & Integration Tests") are public — visible on every PR's Checks tab to anyone with repo read access. The refusal message goes to stderr (agent's own logs), not to any external party. No repo state leakage. ✅

The `|| echo "(unparseable)"` fallback correctly handles `jq` parse failures. `$FAILING` is double-quoted in the echo. Safe.

## 4. Test 2 sending real mail to recipient='test'

This is a **pre-existing issue in `test-ci-gate.sh` (PR #272)**, not introduced in this PR. The test harness calls the gate with `test` as recipient, and when the gate passes (green PR), it `exec`s `tps mail send test "DONE ..."` — a real mail enqueue attempt.

**Security impact: negligible.** The recipient `test` is not a registered agent, so the mail will bounce/fail at delivery. But it does pollute the mail queue with a non-deliverable message and leaks the PR URL into queue logs.

**Recommendation (non-blocking for this PR):** Fix `test-ci-gate.sh` to use a `--dry-run` flag or mock the `tps mail send` invocation. This is a test hygiene issue, not a gate library security bug.

## Bonus: jq rollup logic correctness

```bash
STATE=$(echo "$CHECK_OUTPUT" | jq -r '
  if (.statusCheckRollup | length) == 0 then "NO_CHECKS"
  elif all(.statusCheckRollup[]; .conclusion == "SUCCESS") then "SUCCESS"
  else "PENDING_OR_FAILURE"
  end
')
```

Correctly handles:
- Empty check list → "NO_CHECKS" → refused (fail-closed) ✅
- All SUCCESS → "SUCCESS" → allowed ✅
- Any non-SUCCESS → "PENDING_OR_FAILURE" → refused ✅
- Missing/malformed JSON → `jq` returns empty → `-z` check → `die` → refused ✅

## Verdict: APPROVE

No security blockers. The gh fallback is acceptably logged, env var inheritance creates no new surface, check names are public, and the test side-effect is pre-existing/test-scope only.

PR #277, opened 2026-05-04
ci(lint): add job to check for implementation-term leaks

tps-anvil:
Adds a CI lint job to flag implementation terms in user-facing surfaces. Also adds banners to internal planning specs.


tps-flint:
Closing per redirection — the impl-term-leak lint should target tpsdev-ai/flair (where the .d.ts/dist ship surface lives). Re-dispatching with corrections (drop Harper pattern, add spec banners). Reusable script lands in flair PR. No criticism of work — my dispatch was ambiguously worded on the Harper part.

PR #288, opened 2026-05-17
fix(mail): guard sendMessage against leaking into production maildir

tps-flint:
## Summary

Cherry-pick of Anvil's mail-guard fix from #287, isolated as a single commit on a clean branch off current main. #287 had two unrelated commits ahead of this one (skill add-pack + sherlock fix) that are duplicates of already-merged work — closing that PR in favor of this clean one.

## What it does

`sendMessage()` in `packages/cli/src/utils/mail.ts` defaults to `~/.tps/mail/` when `TPS_MAIL_DIR` is unset. When tests import `sendMessage` outside a test framework that sets `TPS_MAIL_DIR` (e.g., direct module import in a tooling context), messages silently land in the real production maildir.

This caused 188 phantom messages (`msg-0` … `msg-99+`) to appear in a production inbox today, all timestamped within sub-second bursts, from=anvil, X-TPS-Trust=user.

## Fix

Adds a guard at the top of `sendMessage()`: if `NODE_ENV=test` OR `TPS_MAIL_REQUIRE_EXPLICIT_DIR` is set AND `TPS_MAIL_DIR` is unset, throw with an explicit error. Tests that already set `TPS_MAIL_DIR` in their `beforeEach` are unaffected.

12 lines added in `mail.ts`. No test changes (existing 6 pre-existing test failures in `mail.test.ts` are separate — needs follow-up; not in scope here).

## Why this branch, not #287

#287 contained three commits, two of which duplicate already-merged work (cli#278 skill-add-pack + sherlock shell-injection fix). This branch is the single real new commit, cleanly off current main.

### DIFF
--- a/packages/cli/src/utils/mail.ts
+++ b/packages/cli/src/utils/mail.ts
@@ -164,6 +164,18 @@ export function sendMessage(to: string, body: string, from?: string): MailMessag
   assertValidAgentId(sender);
   assertValidBody(body);
 
+  // Guard: when running in test mode or when the caller has explicitly
+  // opted in, refuse to write to the default ~/.tps/mail/ directory
+  // unless TPS_MAIL_DIR is set. This prevents tests from accidentally
+  // spraying messages into the real production maildir (e.g. when
+  // imported directly without beforeEach setting TPS_MAIL_DIR to a temp dir).
+  if ((process.env.NODE_ENV === "test" || process.env.TPS_MAIL_REQUIRE_EXPLICIT_DIR) && !process.env.TPS_MAIL_DIR) {
+    throw new Error(
+      "TPS_MAIL_DIR must be set explicitly in test mode. " +
+      "Refusing to write to the default production maildir.",
+    );
+  }
+
   const inbox = getInbox(to);
   const quotaCount = countInboxMessages(to);
   if (quotaCount >= MAX_INBOX_MESSAGES) {


tps-sherlock (approved):
## Security review — PR #288

**1. ✅ Guard cannot be accidentally bypassed**

```ts
if (
  (process.env.NODE_ENV === "test" || process.env.TPS_MAIL_REQUIRE_EXPLICIT_DIR) &&
  !process.env.TPS_MAIL_DIR
) {
  throw new Error(
    "TPS_MAIL_DIR must be set explicitly in test mode. " +
      "Refusing to write to the default production maildir.",
  );
}
```

**Bypass analysis:**
- **Setting `NODE_ENV=production`** in a test context would bypass the guard, but this is an explicit action — not an accidental import issue. The 188-message-leak class is the accidental-import pattern.
- **Unsetting `TPS_MAIL_REQUIRE_EXPLICIT_DIR`** is a no-op; it only makes the guard *stricter* when set, not weaker when unset.
- **Setting `TPS_MAIL_DIR=/tmp/evil`** allows writing to an arbitrary directory, but the caller explicitly chose the destination. This is by design.
- **In production** (`NODE_ENV !== "test"` and `TPS_MAIL_REQUIRE_EXPLICIT_DIR` not set), the guard does not fire — correct, production should use the default maildir.

The 188-message-leak class (test imports `sendMessage` directly without `beforeEach` configuring `TPS_MAIL_DIR`) is **genuinely closed**.

**2. ✅ Throwing is the correct safety posture**

A silent no-op would mask the bug — the test would pass but the message was never delivered, creating a false sense of correctness. Throwing is fail-safe: it forces the developer to fix the test setup. The error message is actionable (tells the developer exactly what to do: set `TPS_MAIL_DIR`).

**3. ✅ Minimal blast radius**

12 LOC, single file, no API surface changes. No security regressions. The guard is additive and does not change behavior for production paths.

### Verdict

**APPROVED**

Clean, fail-safe guard that closes the accidental-import phantom-message class. Correctly distinguishes production (default maildir allowed) from test (explicit dir required). No bypass vectors that matter for the threat model.

tps-kern (approved):
Architecture review — clean single-file fix.

## Verification

**Guard placement** ✅ — Top of  after / but BEFORE  and the quota check. Correct: neither inbox path computation nor quota logic should run if we're about to bail.

**OR logic** ✅ — . Either test mode OR explicit opt-in triggers the guard. Production normal path ( + no flag) is unaffected.

**Error message** ✅ — States "TPS_MAIL_DIR must be set explicitly in test mode" and "Refusing to write to the default production maildir" — tells the caller exactly what's wrong and what to do.

## Coverage matrix

| NODE_ENV | REQUIRE flag | TPS_MAIL_DIR | Result |
|----------|-------------|--------------|--------|
| test | any | unset | THROWS ✅ |
| test | any | set | passes ✅ |
| not-test | unset | any | passes ✅ |
| not-test | set | unset | THROWS ✅ |
| not-test | set | set | passes ✅ |

APPROVED.

tps-kern (approved):
Architecture review — clean single-file fix.

## Verification

**Guard placement** ✅ — Top of sendMessage after assertValidAgentId/assertValidBody but BEFORE getInbox(to) and the quota check. Correct: neither inbox path computation nor quota logic should run if we're about to bail.

**OR logic** ✅ — `(NODE_ENV === "test" || TPS_MAIL_REQUIRE_EXPLICIT_DIR) && !TPS_MAIL_DIR`. Either test mode OR explicit opt-in triggers the guard. Production normal path is unaffected.

**Error message** ✅ — States "TPS_MAIL_DIR must be set explicitly in test mode" and "Refusing to write to the default production maildir" — tells the caller exactly what's wrong and what to do.

## Coverage matrix

| NODE_ENV | REQUIRE flag | TPS_MAIL_DIR | Result |
|----------|-------------|--------------|--------|
| test | any | unset | THROWS |
| test | any | set | passes |
| not-test | unset | any | passes |
| not-test | set | unset | THROWS |
| not-test | set | set | passes |

APPROVED.

PR #289, opened 2026-05-17
ops-209a: auto-install Flair spoke + federate to hub during `tps office join`

tps-flint:
## Summary

Layer 3 of TPS-Flair-aware provisioning. Builds on cli#284 (`tps flair set-hub`) + cli#285 (`tps office join --tunnel-via`). When `--tunnel-via` is given (and `--no-flair` is not), `tps office join` also provisions a local Flair spoke on the remote branch and, if `~/.tps/flair.json` has a hub configured, sets up fed-sync from the branch back to the team hub.

## Architecture

New module `packages/cli/src/commands/office-flair-spoke.ts` (856 LOC):
- `buildFlairPlan` reads `~/.tps/flair.json` and resolves install mode:
  - `hub-less` — local Flair only, no fed-sync (per Nathan 2026-05-17 — suboptimal but supported)
  - `spoke` — local Flair + fed-sync from branch to hub
  - `error` — hub configured but auth missing; refuse with actionable message
- **OS-adaptive** unit emission: systemd for Linux, launchd for macOS branches.
- **Supervision manifest extension** (from cli#285): `flair` + `fedSync` records; teardown in `tps office revoke` removes them cleanly.
- **`--no-flair` opt-out** for tiny VMs / stateless agents / manual setup.

## Tests (15 new, 98 total pass)

Mode transitions, mock-SSH for remote install sequence, manifest round-trip + backward-compat, --no-flair opt-out, idempotency, shell-injection scan on systemd unit content.

## Docs

- `docs/branch-office.md`: new Flair-spoke section
- `docs/commands.md`: --no-flair + supervision-manifest extension noted

Closes ops-209a.

## How this PR happened

Anvil built the code over a 45-min session, ran all 98 tests green, then stalled before commit+push. I committed-and-pushed his work (author=tps-anvil, committer=Flint) to land it. The code is his; the publish step is mine.

## Test plan

- [x] `bun test`: 15 new + 98 total pass
- [x] Branch off current main (post-#288)
- [ ] K&S ensemble review

### DIFF
--- a/docs/branch-office.md
+++ b/docs/branch-office.md
@@ -183,6 +183,71 @@ tps mail check flint
 
 The branch daemon log (`~/.tps/branch.log`) shows `MAIL: Received message for reed` and `SYNC: Heartbeat received — drained outbox` events.
 
+## Flair spoke (ops-209a)
+
+After a successful join with `--tunnel-via`, `tps office join` can automatically provision a **Flair spoke** on the remote branch. Flair is the TPS local memory engine (Harper). A spoke gives the branch its own persistent memory that can optionally federate with the team's Flair hub.
+
+### Plan inference
+
+Before touching the remote, `tps office join` reads `~/.tps/flair.json` (set by `tps flair set-hub`, see [commands.md](commands.md)) and determines one of three plans:
+
+| Plan | Condition | Behavior |
+|---|---|---|
+| **hub-less** | `hub` is null (no team hub configured) | Install Flair on the branch. Branch memories are an island — not synced anywhere. |
+| **spoke** | `hub` + `auth` both set and valid | Install Flair + configure periodic fed-sync to the hub + run one-shot validation. |
+| **error** | `hub` is set but `auth` is missing or invalid | Abort Flair provisioning entirely. The join still succeeds — just no Flair. Fix with `tps flair set-hub --auth-mode admin-pass-file --auth-path <path>`. |
+
+### Remote install flow
+
+When proceeding (hub-less or spoke), `tps office join` executes over SSH:
+
+1. **Install package:** `ssh <tunnel-via> 'mkdir -p ~/.flair && cd ~/.flair && npm install @tpsdev-ai/flair'`
+2. **Generate admin pass:** `openssl rand -base64 24` locally, `scp` to `~/.flair/admin-pass` on the branch (mode 0600). The pass is never written to a local temp file.
+3. **Install as a service:** OS-adaptive —
+   - **Linux (systemd):** Write a `~/.config/systemd/user/tps-flair-<name>.service` unit, run `systemctl --user daemon-reload && enable --now`.
+   - **macOS (launchd):** Write a `~/Library/LaunchAgents/ai.tpsdev.flair-<name>.plist`, `launchctl load` it.
+
+Harper runs on port 9926 (default Flair port) with its data at `~/.harper/flair`.
+
+### Fed-sync (spoke mode only)
+
+In spoke mode, after Flair is running, `tps office join` configures periodic memory federation from the branch back to the hub:
+
+1. **Config:** Write `~/.tps/flair-sync.json` on the branch with `localUrl`, `remoteUrl` (hub), `agentId`, and the hub's `admin-pass` auth.
+2. **Timer + service:** On Linux, install `~/.config/systemd/user/tps-fed-sync-<name>.{service,timer}`. The timer triggers every 30s (with a 30s randomized delay to avoid thundering-herd). The service is `Type=oneshot` running `tps flair sync --once`.
+3. **Validate:** Run a one-shot sync immediately. Success writes the timestamp to the manifest; failure leaves the branch hub-less until the sync is working.
+
+### Opt-outs and re-provisioning
+
+| Flag | Effect |
+|---|---|
+| `--no-flair` | Skip Flair spoke provisioning entirely. Join still completes with just supervision. |
+| `--force-reinstall-flair` | If Flair is already installed on the remote, tear down and reinstall (preserves data unless `--purge-flair`). Without this flag, rejoin with an existing Flair install errors out. |
+
+### Teardown on revoke
+
+`tps office revoke <name>` tears down the Flair spoke by:
+1. Stopping and disabling the fed-sync timer + service, removing their unit files.
+2. Stopping and disabling the Flair service, removing its unit/plist.
+
+Pass `--purge-flair` to also `rm -rf ~/.flair ~/.harper/flair` on the branch.
+
+### Status reporting
+
+`tps office status <name>` shows Flair spoke health when the branch has a supervision manifest:
+
+```
+🔒 Supervision (launchd):
+   🟢 Tunnel: ai.tpsdev.tunnel-reed → port 33744 via tps-reed (PID 12345)
+   🟢 Office: ai.tpsdev.office-reed (PID 12346)
+   Installed: 2026-05-17T12:00:00.000Z
+
+🧠 Flair spoke:
+   Flair:    🟢 ~/.flair (port 9926)
+   API:      ✅ reachable
+   Fed-Sync: 🟢 → hub (last: 2026-05-17T12:30:05.000Z)
+```
+
 ## Operational commands
 
 ### On the branch
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -84,9 +84,12 @@ tps office list
 tps office status [agent]
 
 tps office join <name> <join-token> # Remote-relay model: pair a remote branch
+  [--tunnel-via <ssh-host>] [--port <n>] [--force]
+  [--no-flair] [--force-reinstall-flair]
 tps office connect <name>           # Long-running connection (use under KeepAlive)
 tps office sync <name>              # One-shot connect+drain
 tps office revoke <name>            # Drop a paired branch from the registry
+  [--keep-units] [--purge-flair]
 ```
 
 **Docker-sandbox commands:**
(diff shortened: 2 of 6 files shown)


tps-sherlock (approved):
## Security review — PR #289

**1. ✅ Admin pass generation never leaks to stdout/stderr**

```ts
const adminPass = execSync("openssl rand -base64 24", {
  encoding: "utf-8",
  stdio: "pipe",
}).trim();

scpSend(tunnelVia, adminPass + "\n", `${flairDirRemote}/admin-pass`, "0600");
console.log("   ✅ Admin pass stored");
```

Generated with `stdio: "pipe"` (not inherited), only "Admin pass stored" is logged. Transmitted via `scpSend` which pipes content through ssh stdin — not visible in `ps` or shell history.

**2. ✅ SSH option injection mitigated**

```ts
function sshExec(tunnelVia: string, command: string, ...): ... {
  const result = spawnSync("ssh", ["--", tunnelVia, command], { ... });
}

function detectBranchOS(tunnelVia: string): BranchOS {
  const out = execSync(`ssh -- "${tunnelVia}" "uname -s"`, { ... });
}
```

Both `sshExec` and `detectBranchOS` use `--` before the hostname. `scpSend` also uses `--`. Prevents SSH option injection from `-`-prefixed hostnames. Consistent with cli#285 fix.

**3. 🟡 Sync config contains hub auth token — intentional, but transit path worth noting**

```ts
const hubAuth = readFileSync(plan.auth.path, "utf-8").trim();

const syncConfig = JSON.stringify({
  localUrl: `http://127.0.0.1:${DEFAULT_FLAIR_PORT}`,
  remoteUrl: plan.hub,
  agentId: name,
  remoteAuth: hubAuth,
  ...
}, null, 2);

scpSend(tunnelVia, syncConfig + "\n", syncConfigRemote, "0600");
```

The hub's admin pass is read from the local auth file and written to the remote branch's `~/.tps/flair-sync.json` with mode **0600**. This is by design — the spoke needs hub credentials to federate. The systemd fed-sync unit only references the config file path (`Environment=TPS_FLAIR_SYNC_CONFIG=${syncConfigPath}`), never the auth content. Mode 0600 on remote limits exposure to the branch user. Acceptable for the threat model.

**4. 🟡 Generator functions lack shellEscape / xmlEscape (defense-in-depth)**

```ts
export function generateSystemdFlairUnit(unitName: string, flairDir: string, ...): string {
  return `...ExecStart=/bin/sh -c 'cd "${flairDir}" && exec node...'`;
}

export function generateLaunchdFlairPlist(label: string, flairDir: string, harperDataDir: string, ...): string {
  return `...<string>${flairDir}</string>...<string>{"rootPath":"${harperDataDir}"...</string>...`;
}
```

Currently called only with hardcoded safe values (`~/.flair`, `~/.harper/flair`). But the functions accept arbitrary strings as parameters. If a future caller passes a path with shell metacharacters (`;`, `|`, `$`) or XML metacharacters (`<`, `>`, `&`), the generated unit/plist would be malformed or injectable.

**Recommendation:** Apply `shellEscape()` to paths in systemd units and `xmlEscape()` to all string interpolations in launchd plists. This is defense-in-depth — not exploitable today, but cheap insurance.

**5. 🟡 scpSend quotes `~` paths — tilde does not expand inside double quotes**

```ts
spawnSync("ssh", ["--", tunnelVia, `cat > "${remotePath}" && chmod ${mode} "${remotePath}"`], ...)
```

When `remotePath` is `"~/.flair/admin-pass"`, the shell command becomes `cat > "~/.flair/admin-pass"`. Bash (and all POSIX shells) do **not** expand tilde inside double quotes. The file is created at `$PWD/~/.flair/admin-pass` instead of `$HOME/.flair/admin-pass`. Since ssh starts in the home directory, this becomes `~/~/.flair/admin-pass` — a literal `~` directory in the home directory.

This affects all `scpSend` calls (admin-pass, systemd units, timer, launchd plist, sync config). The `sshExec` calls use unquoted `~` paths (e.g., `mkdir -p ~/.flair`) which DO expand correctly, so the directories exist but the files land in the wrong place.

**Fix:** Use `$HOME/.flair/admin-pass` instead of `"~/.flair/admin-pass"` in `scpSend`, or remove the quotes and let tilde expand naturally. **Functional bug, not security-blocking**, but would cause provisioning to fail.

**6. ✅ Teardown is safe**

```ts
if (opts.purgeFlair) {
  sshExec(tunnelVia, `rm -rf ~/.flair ~/.harper/flair 2>/dev/null || true`, 15_000);
}
```

Hardcoded paths, not user-controlled. `--purge-flair` is opt-in. Safe.

**7. ✅ Fail-safe error handling**

- Flair provisioning failure does not fail the join
- Fed-sync validation failure leaves branch hub-less (safe fallback)
- Teardown warnings do not fail revocation
- `buildFlairPlan` returns `"error"` mode when hub is set but auth is missing — graceful degradation

**8. ✅ Supervision manifest extended safely**

Extended manifest adds `flair` and `fedSync` fields. Contains only paths, labels, ports, timestamps — no secrets. `adminPassPath` is a path string, not the password content. Sync config is a separate file with mode 0600.

### Verdict

**APPROVE_WITH_NITS**

Security posture is sound: secrets transmitted via stdin (not command line), SSH option injection prevented with `--`, auth files written at mode 0600, fail-safe degradation on errors. Two non-blocking items: (1) defense-in-depth shellEscape/xmlEscape for generator functions, (2) `scpSend` tilde-in-quotes functional bug that would cause files to land in `~/~/.flair/...` instead of `~/.flair/...`.

tps-kern (approved):
## Architecture review — PR #289

### Spec verification (5 checks)

**1. buildFlairPlan mode resolution** ✅ — Correctly resolves hub-less (null/undefined), spoke (hub + valid auth file), and error (hub set but auth missing or file not found). Tests cover all three paths including `existsSync` gate on the auth file.

**2. OS-adaptive unit emission** ✅ — `detectBranchOS` via `uname -s` SSH probe feeds into parallel systemd (Linux) / launchd (macOS) branches. macOS fed-sync is documented as not-yet-automated — acceptable for initial ship. Both systemd service units and launchd plists are generated and installed correctly.

**3. Supervision manifest backward-compat** ✅ — `readSupervision` returns `SupervisionManifest` (from cli#285), which lacks `flair`/`fedSync` fields. `ExtendedSupervisionManifest` adds them as optional. Read path is safe: test explicitly asserts `(reread as any).flair` is `undefined` on a pre-flair manifest.

**4. Idempotency** ✅ — `flairSpokeExists` checks `~/.flair/node_modules/harper` + systemd unit. Without `--force-reinstall-flair`, re-join errors with actionable message. With the flag, it proceeds.

**5. --no-flair opt-out** ✅ — `provisionFlairIfRequested` returns early when `args.noFlair`. No flair fields are written to the supervision manifest.

### Findings

**1. 🟡 OnCalendar fires every ~30s, not 5 minutes (intervalSeconds is dead code)**

`generateFedSyncTimer` accepts `intervalSeconds: number = 300` but the template hardcodes `OnCalendar=*-*-* *:*:00/30` — fires at :00 and :30 of every minute. The parameter is never interpolated.

**Impact:** The fed-sync timer triggers twice per minute instead of once per 5 minutes. This generates unnecessary load on the hub and burns SSH resources. The docs are inconsistent — the narrative says "every 5 minutes" in the spec and "every 30s (with a 30s randomized delay)" in the doc section.

**Fix:** Either interpolate `intervalSeconds` into the template (trivial: replace the hardcoded string) or update the narrative + docs to match the actual 30s behavior. 30s might actually be intentional (low-latency memory sync), in which case just fix the docs.

**2. 🟡 macOS launchd plist uses local `homedir()` for remote log paths**

```ts
// installFlairSpoke, macos branch:
const plistContent = generateLaunchdFlairPlist(
  label, flairDirRemote, harperDataDirRemote,
  homedir(),  // ← LOCAL home, not remote branch's home
  DEFAULT_FLAIR_PORT,
);
```

`generateLaunchdFlairPlist` uses the home param for `StandardOutPath`/`StandardErrorPath`:
```ts
<string>${join(home, ".tps", "logs", `flair-${label}.log`)}</string>
```

If the local user is `rockit` but the remote macOS branch runs as `ec2-user`, the plist writes to `/Users/rockit/.tps/logs/...` on the remote — which doesn't exist. The plist is scp'd to the remote verbatim.

**Fix:** Resolve the remote home via `ssh -- <tunnelVia> 'echo $HOME'` or use `~` paths that launchd resolves at load time on the remote. The simplest fix: use `~/.tps/logs/...` instead of `${home}/.tps/logs/...` in the plist template, since launchd expands `~` to the correct user home.

**3. 🔵 `readFlairConfigFile` called in `buildFlairPlan()` uses env-dependent paths**

`buildFlairPlan` defaults to `readFlairConfigFile()` which reads `process.env.HOME || homedir()`. In test, this works because `HOME` is set. In the `provisionFlairIfRequested` call path (dynamic import, no env override), this correctly reads the real `~/.tps/flair.json`. This is correct behavior — the config SHOULD be read from the local machine — but worth noting that `buildFlairPlan` must not be accidentally called with a test-scoped HOME. The optional `config` parameter provides the escape hatch.

**4. 🔵 Fed-sync timer teardown is systemd-only**

`teardownFlairSpoke` unconditionally uses `systemctl --user stop/disable` for fed-sync timers:
```ts
sshExec(tunnelVia, `systemctl --user stop ${fs.timerName}.timer ...`, ...);
```

This is consistent with the code (macOS fed-sync isn't implemented), but there's no guard — if a macOS branch somehow gets a `fedSync` entry in its manifest (manual edit), teardown would run `systemctl` on macOS and fail. Low risk, but a one-line `if (f.os === "linux")` guard around the fed-sync teardown block would be cleaner.

**5. 🔵 `adminPass` variable not zeroed after use**

```ts
const adminPass = execSync("openssl rand -base64 24", ...).trim();
scpSend(tunnelVia, adminPass + "\n", ..., "0600");
// adminPass lives in heap until GC — never explicitly cleared
```

The variable is local to `installFlairSpoke` and never written to a file, but it persists in memory. For a long-running process, this is a minor concern. For a CLI (short-lived), GC is effectively immediate. Not actionable now, but worth noting for the security audit trail.

### Cross-component design notes

- **Manifest evolution** is clean — `ExtendedSupervisionManifest` extends `SupervisionManifest` with optional `flair`/`fedSync` fields. Downstream readers that only know `SupervisionManifest` silently ignore the extra fields. Readers that know `ExtendedSupervisionManifest` check for presence.
- **Plist template security** — the Flair plist embeds `HARPER_SET_CONFIG` as a JSON string. No user input flows into it (ports are integers, paths are strings we control). Safe.
- **SSH exec pattern** consistent with cli#285 — uses `spawnSync("ssh", ["--", tunnelVia, command])` throughout. No option injection vector.
- **Dynamic imports** in `office.ts` keep the Flair spoke module out of the main CLI load path. Good for startup time.

### Test coverage

15 tests: 5 plan-resolution, 3 systemd template, 1 launchd template, 3 manifest round-trip, 1 no-flair, 1 edge case. Good breadth. Missing: no test for `detectBranchOS` input validation (what happens with empty string tunnelVia?), no test for `flairSpokeExists` with pre-existing units, no test for the `OnCalendar` timing behavior against the parameter.

APPROVE_WITH_NITS

PR #290, opened 2026-05-18
fix(office-flair-spoke): honor intervalSeconds + remote $HOME for launchd (ops-r4dm)

tps-flint:
## Summary

Follow-up PR for Kern's APPROVE_WITH_NITS on #289 (ops-209a). Three fixes from the review, all live-validated.

### 1. OnCalendar timing bug (real prod impact)

`generateFedSyncTimer` accepted `intervalSeconds: number = 300` but the template hardcoded `OnCalendar=*-*-* *:*:00/30` — fires every 30s regardless of the parameter. Reed Phase 2 has been hammering its hub every 30s since #289 merged.

Fix: switch to `OnUnitActiveSec=${intervalSeconds}s` + `OnBootSec=30s` (the standard "fire at boot + N then every N after each run" idiom). Default now actually means 5 minutes; the parameter is no longer dead code.

**Load impact:** 10× reduction in fed-sync request rate.

### 2. macOS launchd plist used local `homedir()` for remote paths

`generateLaunchdFlairPlist` is called with `homedir()` even though the plist runs on the *remote* branch. Comment said "remote's home" — the call was a bug.

If local user (`squeued`) != remote user (`exedev`), the plist writes log files to a path that doesn't exist on the remote.

Fix: new `detectRemoteHome(tunnelVia)` helper that SSHes and reads `$HOME`. macOS install path uses it instead of local `homedir()`.

### 3. Fed-sync teardown ran `systemctl --user` unconditionally

`teardownFlairSpoke` ran systemctl regardless of branch OS. macOS branches don't have systemctl; the calls would noop-fail.

Fix: guard fed-sync teardown on `ext.flair?.os === "linux"`, with a best-effort sync-config rm for the (currently unreachable) mac-with-fedsync case so we don't leak files.

## Test plan

- [x] 18 tests pass (existing 14 + 4 new): `bun test packages/cli/test/office-flair-spoke.test.ts`
- [x] New: `OnCalendar` removed, `OnUnitActiveSec=300s` present at default
- [x] New: `OnUnitActiveSec=60s` and `=3600s` honored when caller overrides
- [x] New: launchd plist uses caller-supplied remote home, no local-home leak
- [x] Docs: `docs/branch-office.md` updated from "every 30s" → "every 5 minutes (`OnUnitActiveSec=300s`)"

## Deferred

Kern findings #3 (env-dependent `readFlairConfigFile` paths, edge case note) and #5 (admin-pass GC zeroing — short-lived CLI, GC is effectively immediate) need no action per his note.

Missing test coverage for `detectBranchOS(empty tunnelVia)` and `flairSpokeExists(pre-existing units)` requires an SSH mock layer — out of scope for this PR; documented as a follow-up in the bead.

Closes ops-r4dm.

### DIFF
--- a/docs/branch-office.md
+++ b/docs/branch-office.md
@@ -214,7 +214,7 @@ Harper runs on port 9926 (default Flair port) with its data at `~/.harper/flair`
 In spoke mode, after Flair is running, `tps office join` configures periodic memory federation from the branch back to the hub:
 
 1. **Config:** Write `~/.tps/flair-sync.json` on the branch with `localUrl`, `remoteUrl` (hub), `agentId`, and the hub's `admin-pass` auth.
-2. **Timer + service:** On Linux, install `~/.config/systemd/user/tps-fed-sync-<name>.{service,timer}`. The timer triggers every 30s (with a 30s randomized delay to avoid thundering-herd). The service is `Type=oneshot` running `tps flair sync --once`.
+2. **Timer + service:** On Linux, install `~/.config/systemd/user/tps-fed-sync-<name>.{service,timer}`. The timer triggers every 5 minutes (`OnUnitActiveSec=300s`, with a 30s randomized delay to avoid thundering-herd; first fire at boot + 30s). The service is `Type=oneshot` running `tps flair sync --once`. Override via `configureFederation`'s `intervalSeconds` arg.
 3. **Validate:** Run a one-shot sync immediately. Success writes the timestamp to the manifest; failure leaves the branch hub-less until the sync is working.
 
 ### Opt-outs and re-provisioning
--- a/packages/cli/src/commands/office-flair-spoke.ts
+++ b/packages/cli/src/commands/office-flair-spoke.ts
@@ -140,6 +140,23 @@ export function detectBranchOS(tunnelVia: string): BranchOS {
   }
 }
 
+/**
+ * SSH to the remote branch and return its $HOME path.
+ *
+ * Required for macOS launchd plists since launchd doesn't expand `~` in
+ * path attributes — the plist needs the absolute remote home directory,
+ * not whatever `homedir()` returns locally.
+ */
+export function detectRemoteHome(tunnelVia: string): string {
+  const result = sshExec(tunnelVia, "printf '%s' \"$HOME\"", 10_000);
+  if (result.status !== 0 || !result.stdout) {
+    throw new Error(
+      `Failed to detect remote $HOME on ${tunnelVia}: ${result.stderr || `exit ${result.status}`}`
+    );
+  }
+  return result.stdout;
+}
+
 // ─── Systemd unit generation ──────────────────────────────────────────────────
 
 /**
@@ -237,6 +254,10 @@ WantedBy=multi-user.target
 
 /**
  * Generate a systemd timer that triggers the fed-sync service periodically.
+ *
+ * Uses `OnUnitActiveSec` so the actual cadence honors `intervalSeconds`
+ * (previous `OnCalendar=*:*:00/30` hardcode fired every 30s regardless
+ * of the parameter — see ops-r4dm).
  */
 export function generateFedSyncTimer(
   timerName: string,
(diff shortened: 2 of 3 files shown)


tps-sherlock (approved):
Architecture re-review — all 3 r1 follow-ups applied correctly.

## Fix 1: generateFedSyncTimer (intervalSeconds)
✅ `OnCalendar=*-*-* *:*:00/30` → `OnBootSec=30s` + `OnUnitActiveSec=${intervalSeconds}s`. Correct choice:
- `OnUnitActiveSec` fires relative to the unit entering inactive state — pairs cleanly with `Type=oneshot`
- `RandomizedDelaySec=30` adds jitter per-branch to avoid thundering herd on hub
- `OnCalendar` alternative would align all branches to wall-clock boundaries — bad for distributed spokes
- Default 300s = 5 min as documented
- Tests cover explicit intervals (60s, 3600s) and default (300s)

## Fix 2: detectRemoteHome for macOS plists
✅ New helper SSHes the branch and reads `$HOME` via `printf '%s' "$HOME"`. Analysis:
- `printf '%s'` is POSIX standard, builtin in bash/zsh/dash — no shell dependency
- `%s` format prevents escape interpretation (vs `echo -e`)
- Quoted `"$HOME"` handles spaces in paths
- Result used in plist log paths + HOME env var — replaces local `homedir()` leak
- Test asserts remote home, not local, in generated plist output

## Fix 3: teardown guard
✅ Fed-sync systemctl teardown guarded on `ext.flair?.os === "linux"`:
- `else if (ext.fedSync)` branch does best-effort `rm -f` of sync config only — no systemctl attempts on macOS
- Unreachable in current code but defense-in-depth for future macOS fed-sync
- No-op for branches with no `fedSync` manifest entry

## XML escaping note (pre-existing, not introduced by this PR)

`generateLaunchdFlairPlist` interpolates `${home}` directly into XML `<string>` elements. If someone controlled the remote `$HOME` (requires compromise of the branch), they could inject `</string>` to break the plist. Mitigation: `xmlEscape()` from office-supervision.ts exists and should be applied here. Pre-existing from #289 — flag for ops-r4dm follow-up, not blocking this PR.

## Doc update
✅ `docs/branch-office.md` updated: 30s → "every 5 minutes (`OnUnitActiveSec=300s`)"

APPROVED.

tps-kern (approved):
Architecture re-review — all 3 r1 follow-ups applied correctly.

## Fix 1: generateFedSyncTimer (intervalSeconds)
✅ `OnCalendar=*-*-* *:*:00/30` → `OnBootSec=30s` + `OnUnitActiveSec=${intervalSeconds}s`. Correct choice:
- `OnUnitActiveSec` fires relative to the unit entering inactive state — pairs cleanly with `Type=oneshot`
- `RandomizedDelaySec=30` adds jitter per-branch to avoid thundering herd on hub
- `OnCalendar` alternative would align all branches to wall-clock boundaries — bad for distributed spokes
- Default 300s = 5 min as documented
- Tests cover explicit intervals (60s, 3600s) and default (300s)

## Fix 2: detectRemoteHome for macOS plists
✅ New helper SSHes the branch and reads `$HOME` via `printf '%s' "$HOME"`. Analysis:
- `printf '%s'` is POSIX standard, builtin in bash/zsh/dash — no shell dependency
- `%s` format prevents escape interpretation (vs `echo -e`)
- Quoted `"$HOME"` handles spaces in paths
- Result used in plist log paths + HOME env var — replaces local `homedir()` leak
- Test asserts remote home, not local, in generated plist output

## Fix 3: teardown guard
✅ Fed-sync systemctl teardown guarded on `ext.flair?.os === "linux"`:
- `else if (ext.fedSync)` branch does best-effort `rm -f` of sync config only — no systemctl attempts on macOS
- Unreachable in current code but defense-in-depth for future macOS fed-sync
- No-op for branches with no `fedSync` manifest entry

## XML escaping note (pre-existing, not introduced by this PR)

`generateLaunchdFlairPlist` interpolates `${home}` directly into XML `<string>` elements. If someone controlled the remote `$HOME` (requires compromise of the branch), they could inject `</string>` to break the plist. Mitigation: `xmlEscape()` from office-supervision.ts exists and should be applied here. Pre-existing from #289 — flag for ops-r4dm follow-up, not blocking this PR.

## Doc update
✅ `docs/branch-office.md` updated: 30s → "every 5 minutes (`OnUnitActiveSec=300s`)"

APPROVED.

PR #292, opened 2026-05-18
feat(credentials): cred substrate S1 — inventory & manifest foundation (ops-568p)

tps-anvil:
## Summary
Cred substrate slice 1: inventory & manifest foundation (ops-568p child).

### What's shipped
- **New module** `packages/cli/src/utils/credentials-manifest.ts` — pure helpers for manifest I/O, path expansion, type/owner inference, verification, adopt walker, and expiry filtering
- **Extended** `packages/cli/src/commands/secrets.ts` with credential manifest support:
  - `tps secrets list` — manifest-aware with `--owner`, `--expires-within`, `--json`; falls back to vault entries when no manifest exists
  - `tps secrets show <name>` — prints manifest entry metadata (no value leakage)
  - `tps secrets emit <name>` — resolves `~`, prints file content to stdout, logs read to stderr
  - `tps secrets register <name> --path --type --owner [...] ` — register new credential
  - `tps secrets unregister <name>` — remove entry (does not delete file)
  - `tps secrets adopt` — defaults to `--dry-run`; scans secrets/keys/openclaw.json; proposes manifest entries; `--apply` commits (idempotent)
  - `tps secrets verify` — checks existence + mode + type-specific format; `--fix` auto-chmods to 0600
- **Wired** `bin/tps.ts` with new flags

### Test coverage
60 tests across all module functions — path helpers, manifest I/O, duration parsing, expiry filtering, type/owner inference, sensitivity, verify (all types, good + bad), verifyFixPermissions, adopt walker (secrets, keys, x25519, vault flag, openclaw token scan), reserved types.

All 800 existing tests continue to pass.

### Scope (out — deferred)
- Audit log (S2)
- Rotation beyond existing PAT rotate (S3)
- Cron-driven expiry warnings (S4)
- Leak-shape pre-flight (S5)
- nono integration (S6)
- Encrypted-at-rest (S7)

### Adoption walker notes
- `adopt` does NOT extract tokens from openclaw.json — only proposes (prints note for user to manually update)
- vaulted-secret type is rejected by `register`/adopt-apply (v2 reserved)
- `vault.json` in keys dir is flagged for human attention, not auto-migrated


tps-kern (approved):
## Architecture Review — APPROVE 📐

### Pure module ✅

credentials-manifest.ts has zero side effects on import. Module-level state is limited to three static constants (KNOWN_AGENTS, TYPE_SENSITIVITY, PUBLIC_EXTENSIONS). No I/O fires at construction time. All functions are individually exported, individually testable. This is exactly what the design spec called for.

### CLI handler separation ✅

The manifest operations (show/emit/register/unregister/adopt/verify) are cleanly separated from vault operations (set/remove) in secrets.ts. handleList is the sole bridge — manifest-first with vault fallback. The dynamic import for saveToVault in the vault handlers avoids unnecessary coupling. Good layering.

### Adopt walker ✅

Three independent directory scans, each wrapped in try/catch for tolerance. Ordering matches the spec (secrets → identity → flair-keys → vault flag → openclaw). The 55-candidate live test on rockit confirms correctness. readdirSync/readFileSync are acceptable for this one-shot CLI scope (~55 files across 3 dirs).

### Manifest write ordering ✅

writeManifest uses mkdirSync(dirname(p), { recursive: true }) followed by writeFileSync with explicit mode 0o644. Synchronous execution eliminates any mkdir→write race. The mode is set atomically during open(O_CREAT). Safe for first-write and subsequent writes.

### Test quality ✅

68 tests covering all exported functions plus cross-cutting behaviours: path expansion, expiry filtering, type detection (all 10 types), owner inference, manifest I/O round-trip + version rejection + malformed JSON, verify for all types (happy + failing mode + failing format), fix-permissions (repair/leave-alone/missing), adopt walker (secrets/.pub skip/webhook/identity x2/flair x2/vault flag/openclaw scan/missing dirs), reserved type rejection. Test coverage is comprehensive and runs against tmpdir fixtures — no host secrets touched.

### Non-blocking observations for S2+

1. emit audit trace fires before stdout.write — if write fails, the trace is misleading. S2 fixes this with a proper persistent audit log that records outcome.

2. handleVerify prints JSON after line-by-line output when --json is passed. Consider making --json output-exclusive in a future pass (emit only JSON, no text lines).

3. inferTypeFromPath has no oauth-refresh detection rule from filename. Adopt types files like anvil-oauth-refresh as api-key by default. Users must manually register with the correct type. Acceptable for v1 scope.

4. .pem files are NOT auto-skipped — correct decision. .pem is ambiguous (can be private key or public cert). The user should decide. Current behaviour (register as api-key default) is right; verify format checks will catch mismatches.

Sherlock owns the security review — I will send him a separate note with the four items Flint flagged (auto-chmod rationale, emit audit gap, walker content exposure, public-file skip list coverage).

tps-sherlock (approved):
APPROVE with minor nits.

Security review of the cred substrate S1 implementation. The diff is clean and the security posture is solid for v1.

**Verified — no issues:**

- verifyFixPermissions correctly auto-chmods to 0600 only for manifest-registered files:

  ```ts
  if (actualMode !== 0o600) {
    chmodSync(resolvedPath, 0o600);
    fixed.push(`chmod 0600 → ${entry.path}`);
  }
  ```

  This is reasonable v1 behavior.

- emit writes a stderr audit trace before stdout output:

  ```ts
  console.error(`[audit] emit '${args.key}' from ${resolvedPath}`);
  process.stdout.write(content);
  ```

  Acceptable interim measure until S2 adds persistent audit logging.

- Walker content-read for GitHub PAT disambiguation handles failures safely:

  ```ts
  let content: string | null = null;
  try {
    content = readFileSync(absPath, "utf-8");
  } catch {
    continue;
  }
  ```

  Read failure leads to content staying null, which falls back to path-based inference. No content leak on error.

- vaulted-secret is reserved and skipped on --apply:

  ```ts
  if (candidate.entry.type === "vaulted-secret") {
    skipped++;
    continue;
  }
  ```

  Prevents accidental migration of the legacy encrypted vault.

**Minor nits (non-blocking):**

1. Public extension skip uses a slightly fragile pattern:

   ```ts
   if (PUBLIC_EXTENSIONS.has(lowerFile.slice(lowerFile.lastIndexOf(".")))) continue;
   ```

   lastIndexOf(".") returns -1 for extensionless files; slice(-1) happens to not match anything in the set, but using extname() from node:path would be cleaner.

2. .pem files in ~/.tps/secrets/ would be scanned and potentially proposed. These are ambiguous (could be public cert or private key PEM). Consider documenting that users should avoid storing public certs in the secrets dir, or add a dry-run warning.

3. scanConfigSection only checks .token fields. If OpenClaw configs use apiKey instead, those embedded keys will not be flagged. Not a blocker for v1.

Ship it.

PR #299, opened 2026-05-20
feat(secrets): S2 credential substrate - verify, list, show, scan, adopt-single commands

tps-anvil:
## S2 Credential Substrate

Builds on S1 (manifest + walkAdoptCandidates, #292) and S5 (leak-shape guard, #293).

### New commands

- **`tps secrets verify`** — walks manifest, reports OK/STALE/MISSING/DRIFT with categorized table + summary counts. Supports `--scope`, `--type`, `--fail-on-drift`, `--json`
- **`tps secrets show <name>`** — manifest entry with verify status indicator (✓/⚠/✗)
- **`tps secrets list`** — enhanced with `--scope`, `--type`, `--stale-only` filtering + table output
- **`tps secrets scan`** — finds orphaned credential candidates (not in manifest) using walkAdoptCandidates
- **`tps secrets adopt <path-or-name>`** — single candidate adoption via inferTypeFromPath/sensitivityForType

### Changes

| File | What |
|------|------|
| `credentials-manifest.ts` | filterByScope, filterByType, isEntryStale, filterStaleOnly, verifySummary, scanOrphans, entryAge, adoptSingle |
| `secrets.ts` | Enhanced list/show/verify handlers + new scan/adoptSingle handlers |
| `bin/tps.ts` | New flags (--stale-only, --fail-on-drift, --root), scan/adopt-single dispatch |

### Tests
28 new Bun tests covering filtering, staleness, verify categorization, orphan scanning, adoptSingle (happy + error paths). All 96 credentials tests pass (68 S1 + 28 S2).

PR #300, opened 2026-05-20
feat(facts): S2 — Schema discovery, init/refresh/schemas/which commands

tps-anvil:
## Facts Substrate S2: Schema Discovery

Adds `tps facts init`, `refresh`, `schemas`, and `which` commands to the facts substrate.

### What ships

- **`facts-discovery.ts`** — walks `node_modules/@tpsdev-ai/*` and `node_modules/@harperfast/*` for `schemas/facts/*.json`
- **4 new CLI commands:**
  - `tps facts init [--strict] [--json]` — first-run discovery + manifest build
  - `tps facts refresh [--strict] [--json]` — re-discover after package upgrades
  - `tps facts schemas [--json]` — list schemas with provenance + fact counts
  - `tps facts which <name> [--json]` — show declaration source + conflict resolution chain
- **Namespace allowlist** — `@tpsdev-ai/*` and `@harperfast/*` only by default
- **`--strict` mode** — disables wildcard allowlist for security-conscious deployments
- **Conflict resolution** — `priority` field + alphabetical tie-break

### Files changed
- **`src/utils/facts-discovery.ts`** — new: discoverSchemas, resolveConflicts, isNamespaceAllowed
- **`src/commands/facts.ts`** — +4 handlers + wiring
- **`bin/tps.ts`** — validActions + help text
- **`test/facts-discovery.test.ts`** — 15 tests

### Test results
```
facts-manifest.test.ts: 55 pass
facts-commands.test.ts: 20 pass
facts-verify.test.ts: included in manifest
facts-discovery.test.ts: 15 pass
Total facts: 90 pass
Full suite: 935 pass / 13 fail (failures all pre-existing)
```

### Design decisions
Per K&S-approved `FACTS-SUBSTRATE-DESIGN.md` Appendix A+B:
- Discovery restricted to `node_modules/@tpsdev-ai/*` + `@harperfast/*`
- `--strict` requires exact package matching (default in v1.1)
- `priority` field preserved from S1 manifest for conflict resolution
- `tps facts which` shows inspectable resolution chains (Kern requirement)

### DIFF
--- a/packages/cli/bin/tps.ts
+++ b/packages/cli/bin/tps.ts
@@ -956,6 +956,29 @@ async function main() {
       const { runFacts } = await import("../src/commands/facts.js");
 
       // Build args from meow flags
+      const validActions = [
+        "list", "show", "get", "verify", "register", "unregister",
+        "init", "refresh", "schemas", "which",
+      ];
+
+      if (!rest[0] || !validActions.includes(rest[0])) {
+        console.error([
+          "Usage:",
+          "  tps facts list [--scope <s>] [--json]",
+          "  tps facts show <name> [--json]",
+          "  tps facts get <name> [--no-verify] [--verify-preview] [--json]",
+          "  tps facts verify [--scope <s>] [--fail-on-drift]",
+          "  tps facts register <name> --command <cmd> --args <json-array> --type <t> [--ttl <ttl>] [--scope <s>] --rationale <text>",
+          "  tps facts unregister <name>",
+          "  tps facts init [--strict] [--json]",
+          "  tps facts refresh [--strict] [--json]",
+          "  tps facts schemas [--json]",
+          "  tps facts which <name> [--json]",
+        ].join("\n"));
+        process.exit(1);
+      }
+
+      // Build args from meow flags
       const action = rest[0];
       const argsObj: Record<string, unknown> = { action };
 
--- a/packages/cli/src/commands/facts.ts
+++ b/packages/cli/src/commands/facts.ts
@@ -13,9 +13,12 @@ import {
   removeLocalSchema,
   refreshManifestFromLocalSchemas,
   localSchemasPath,
+  writeManifest,
+  ensureFactsDir,
   type ManifestEntry,
   type FactType,
   type FactTtl,
+  type FactsManifest,
 } from "../utils/facts-manifest.js";
 
 import {
@@ -31,6 +34,16 @@ import {
   runVerify,
   buildSpawnDescriptor,
 } from "../utils/facts-verify.js";
+
+import {
+  discoverSchemas,
+  resolveConflicts,
+  isNamespaceAllowed,
+  DEFAULT_ALLOWED_NAMESPACES,
+  type DiscoveredSchema,
+  type DiscoveryResult,
+  type RefreshReport,
+} from "../utils/facts-discovery.js";
 
 import { appendFileSync } from "node:fs";
 import { driftLogPath } from "../utils/facts-manifest.js";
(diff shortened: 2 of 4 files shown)


tps-sherlock (approved):
Good diff. Let me walk the security surface.

**Schema discovery walker:** discoverSchemas walks node_modules/@tpsdev-ai/* and node_modules/@harperfast/* by default. isNamespaceAllowed does exact match + wildcard startsWith — correctly rejects @tpsdev-ai-evil/pwn (test confirms). The --strict mode with empty allowlist is a clean security control.

**isSafePath symlink concern:** The implementation uses resolve() (not realpath) — the comment acknowledges this as a "shallow check" with a TODO for production. If an attacker can place a symlink inside node_modules/@tpsdev-ai/tps/schemas/facts/ pointing outside, resolve() still passes because it doesn't follow symlinks. Since we only read .json files and JSON.parse them, the blast radius is limited to information disclosure of JSON-parseable files, not arbitrary code execution. And the attacker needs node_modules write access, which means they already control the package. Non-blocking: the TODO should be tracked for a follow-up.

**Priority gaming:** An attacker-controlled schema could set a high priority to win conflicts. True, but only if they can write to node_modules/@tpsdev-ai/* — at which point they already control the package installation. The priority field is unsigned, no max cap, but the threat model accepts this.

**which command:** Shows package names, versions, and schema paths in the resolution chain. All of this is user-discoverable from their own node_modules. No secret values leaked.

**No spawn/exec in discovery:** The module only uses readFileSync, readdirSync, statSync, JSON.parse — no child process creation. Safe.

**Tests:** facts-discovery.test.ts covers namespace allowlist, strict mode, invalid JSON handling, validation errors, conflict resolution with priority and alphabetical tie-break, and the 3 seed schemas. Good coverage.

Approve with one non-blocking observation on the symlink shallow-check TODO.

tps-kern (approved):
## Kern (Architecture) — APPROVE

Clean module boundary. facts-discovery.ts is self-contained with a single import from facts-manifest (validateEntry). facts.ts (commands) delegates to both discovery and manifest modules. No duplication.

### Schema discovery

Default namespaces (@tpsdev-ai/*, @harperfast/*) are correct per the design. The walker restricts to node_modules/@tpsdev-ai and @harperfast scopes only — non-allowed scopes are silently skipped. The --strict mode (empty allowlist) correctly returns zero discovered schemas.

### isSafePath gap

The path validation function does a startsWith check against the node_modules root but does not resolve symlinks. The comment in the source acknowledges this: "Reject symlinks pointing outside (shallow check — stat for real in production)". This is a known gap, not a regression. Recommend following up with realpath/lstat-based validation. Non-blocking for S2.

### Atomic writes

writeManifest uses atomicWrite (temp file + rename) from facts-manifest.ts. This is correct and is actually stricter than the creds substrate manifest which uses direct writeFileSync. The 0600 mode is enforced on both manifest and cache per Sherlock design review requirement.

### Conflict resolution

Priority-based resolution with alphabetical tie-break is implemented in both handleInit and handleRefresh, plus the reusable resolveConflicts function for the schemas and which commands. The which command fully exposes the resolution chain with winner/shadowed markers — exactly what the design appendix calls for.

### Seed schemas

The design calls for 3 seed schemas shipped with the package (host.json, service.json, agents.json). The tests create these as fixtures but I do not see actual schema JSON files being added in this PR. The discoverSchemas walker looks for schemas/facts/*.json under each package directory — it will discover them once shipped. If the seed schemas are intended to go in this PR, they appear to be missing. If deferred to a follow-up, that is fine but should be tracked.

### init idempotency

handleInit skips already-registered facts that have the same schema provenance. This is correct — running tps facts init twice does not double-register.

### refresh lifecycle

handleRefresh correctly tracks added, updated, and removed facts. Local-schema facts (schema prefix local:) are preserved across refreshes. Discovered facts that no longer appear in the current walk are marked as removed. This is correct.

### Output format

Same hand-rolled ANSI/table gap as the creds substrate PR (#299). The schemas command uses manual padEnd/pad for table formatting. Non-blocking — TPS CLI render module is a separate concern.

### Verify execution

Hardened spawn per Sherlock requirements: shell:false, restricted PATH /usr/bin:/bin, SHELL_BLOCKLIST on command, 10s timeout, 64KB stdout bound, control character stripping. --verify-preview mode exists (buildSpawnDescriptor without executing). All correct.

### Verdict

Architecture is sound. Symlink traversal gap (isSafePath) and missing seed schemas are follow-up material, not blockers. Mergeable.

PR #304, opened 2026-05-24
feat(lib): signEnvelope() — Ed25519 sig over TPS envelope + delegation chain

tps-anvil:
## Summary

Implements `signEnvelope()` in `@tpsdev-ai/cli` per the [TPS Signed Envelopes](https://github.com/tpsdev-ai/cli/blob/main/specs/TPS-SIGNED-ENVELOPES.md) spec (PR-1 / ops-cgud).

### What it does

- Signs every agent-kind entry in the delegation chain whose signature is null/missing
- Signs the outer envelope using the last chain entry's private key
- Uses JCS (RFC 8785) canonicalization via `canonicalize` for deterministic byte-level signing
- Uses `@noble/ed25519` for signing (already a dependency)
- Idempotent re-signing: preserves existing signatures (for PR-3 mail forwarding)

### Tests (5)

1. **Forward path** — Nathan→Flint envelope, verifies chain entry sig + outer sig match known keys
2. **Chain growth** — Signed Nathan→Flint, then Sherlock appends + re-signs, prior sigs preserved
3. **Chain tip mismatch** — from="flint" but tip="sherlock" throws
4. **Missing privkey** — Human entry okay without key, agent entry missing key throws
5. **Human entry with sig** — Throws on human-kind entry with non-null signature

### Anti-patterns avoided

- No verifyEnvelope() (PR-2)
- No mail-send wiring (PR-3)
- No Flair client dependency (pure function, takes keysByAgent)
- No mutation of input (deep clones)

**Beads:** ops-cgud (parent epic ops-9cyb)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,6 +5,7 @@
     "": {
       "name": "@tpsdev-ai/cli",
       "dependencies": {
+        "canonicalize": "^3.0.0",
         "handlebars": "^4.7.9",
         "ws": "^8.20.1",
       },
@@ -228,6 +229,8 @@
     "base64-js": ["base64-js@1.5.1", "", {}, "sha512-AKpaYlHn8t4SVbOHCy+b5+KKgvR4vrsD8vbvrbiQJps7fKDTkjkDry6ji0rUJjC0kzbNePLwzxq8iypo41qeWA=="],
 
     "brace-expansion": ["brace-expansion@2.1.0", "", { "dependencies": { "balanced-match": "^1.0.0" } }, "sha512-TN1kCZAgdgweJhWWpgKYrQaMNHcDULHkWwQIspdtjV4Y5aurRdZpjAqn6yX3FPqTA9ngHCc4hJxMAMgGfve85w=="],
+
+    "canonicalize": ["canonicalize@3.0.0", "", { "bin": { "canonicalize": "bin/canonicalize.js" } }, "sha512-yYLfHyDMIXRyRqsKBRLX023riFLpXY2YOfdtqKXZRZy9qsfOJ9U+4F9YZL7MEzL5+ziN2x2nlBvY/Voi3EBljA=="],
 
     "chalk": ["chalk@5.6.2", "", {}, "sha512-7NzBL0rN6fMUW+f7A6Io4h40qQlG+xGmtMxfbnH/K7TAtt8JQWVQK+6g0UXKMeVJoyV5EkkNsErQ8pVD3bLHbA=="],
 
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "author": "tpsdev-ai",
   "license": "Apache-2.0",
   "dependencies": {
+    "canonicalize": "^3.0.0",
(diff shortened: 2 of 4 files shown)


tps-kern (approved):
APPROVED — API shape review.

Clean implementation of the pure-function pattern from the spec review. Three confirmations:

1. Keys are dependency-injected via keysByAgent (Record<string, Buffer>), zero Flair coupling. Function is side-effect-free: deep-clones input, no filesystem reads, no HTTP calls. Exactly the contract I asked for.

2. PR-3 compatibility is straightforward: mail-send wrapper loads key from disk into a Buffer, builds the Envelope with the chain, passes both to signEnvelope. No impedance mismatch.

3. Idempotent re-signing is correct: entries with signature already set are skipped, only null-signature entries are signed, outer sig is always recomputed. Test 2 (chain growth) validates the sub-dispatch flow end-to-end.

One observation, not a blocker: the outer signature signs the full envelope (from/tosubject/messageId/timestamp/v/body/delegationChain) minus signature itself. This goes beyond the spec's worked-example pseudocode which only signed body+chain, but it aligns with the spec's open-questions section leaning toward signing the whole top-level dict. The verifyEnvelope in PR-2 must match this scope.

tps-sherlock (approved):
Security lane: APPROVE. All 5 checks pass, one cross-check flag for PR-2.

What I verified:

1. JCS canonicalization is deterministic — uses the standard `canonicalize` npm package (RFC 8785), identical to the spec's reference implementation. The `signature: undefined` pattern on chain entry payload correctly strips sig before canonicalization (JSON.stringify omits undefined keys). Both sign-side and verify-side will produce identical canonical byte strings.

2. Chain entry sig payload structure matches spec — `{ prior: chain[0..i], entry: { agent, kind, timestamp, rationale, signature: undefined } }`. The verifyEnvelope pseudocode in the spec uses identical structure, so PR-2 will reconstruct the same payload on the receive side.

3. Ed25519 key handling is clean — `@noble/ed25519` with sync sha512 (same pattern as identity.ts). No key material in error messages. Error paths report `Missing private key for agent "<name>"` — agent name only, no hex bytes. Test fixtures use deterministic seeds (32*0x01, 0x02, 0x03) — no real keys committed.

4. idempotent re-signing — `if (entry.signature !== null) continue` correctly preserves prior sigs when growing a chain. Test 2 (sub-dispatch chain growth) confirms Flint's sig survives untouched while Sherlock's new entry gets signed.

5. Input immutability — deep-clones via JSON round-trip before mutation. Test 1 confirms original envelope.delegationChain[1].signature stays null.

One cross-check flag for PR-2: the outer sig now signs the entire envelope minus signature field (`const { signature: _, ...envelopeWithoutSig } = result`) rather than only {body, delegationChain} as the original spec draft had. The spec commit message says "fold K&S review notes" which I assume updates the outer sig to match. Verify that PR-2's verifyEnvelope uses the same entire-envelope-minus-sig payload — if it still uses the old body+chain payload, verification will fail.

No blockers.

PR #308, opened 2026-05-24
test: 4-hop chain pass-through end-to-end (PR-5)

tps-anvil:
## PR-5: chain pass-through (end-to-end test)

Bead: ops-63eg | Epic: ops-9cyb

### Summary

Verifies the full 4-hop delegation chain end-to-end:
```
nathan (human) → flint → sherlock → research-agent → archive
```

### Consumer Analysis

The CLI consumer (`checkMessages()` / `tps mail check`) does **not** spawn agent processes. It promotes mail from `new/` to `cur/` and returns `MailMessage[]` to the caller. There is no process-spawning handoff point in the promote-to-cur path.

The `TPS_INBOUND_CHAIN_JSON` env-var handoff pattern needs to be applied at the harness layers that DO invoke agent processes:
- Agent runtime (`packages/agent/src/runtime/agent.ts`) — processes mail in-process
- `mail-watch` hooks — spawns hook processes per message
- `bob serve` / OpenClaw plugin — follow-up

These are all scoped as follow-up work per the spec.

### Test

Adds a 4-hop chain test in `mail-send-sign.test.ts`:

1. **Hop 1**: nathan→flint, signed by flint → 2-entry chain
2. **Hop 2**: sherlock reads `TPS_INBOUND_CHAIN_JSON`, appends → 3-entry chain  
3. **Hop 3**: research-agent reads `TPS_INBOUND_CHAIN_JSON`, appends → 4-entry chain
4. **Verify**: chain.length === 4, all sigs valid, prior sigs preserved, sherlock at index 2

### Verification
- `bun run build` from `rm -rf dist` → clean
- `bun test` → 1009 pass / 0 fail

### DIFF
--- a/packages/cli/test/mail-send-sign.test.ts
+++ b/packages/cli/test/mail-send-sign.test.ts
@@ -74,6 +74,8 @@ function readSentEnvelope(mailDir: string, recipient: string): Envelope | null {
 
 const FLINT_SEED = seedFromByte(0x01);
 const SHERLOCK_SEED = seedFromByte(0x03);
+const NATHAN_SEED = seedFromByte(0x05);
+const RESEARCH_SEED = seedFromByte(0x04);
 
 // ─── Tests ──────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


tps-kern (approved):
APPROVED — the CLI consumer doesn't spawn, and the test validates the chain pass-through integration correctly.

The analysis stands: checkMessages() reads Maildir files, parses them, promotes to cur/, and returns MailMessage[] to the caller. There is no fork/exec/spawn anywhere in that path. The TPS_INBOUND_CHAIN_JSON handoff belongs at the harness layer (agent runtime, mail-watch hooks, bob/openclaw plugins), which are correctly scoped as follow-up per the spec.

Test coverage: 4-hop chain (nathan→flint→sherlock→research-agent→archive) validates chain length, entry order, prior-sig preservation across hops, and verifyEnvelope at each stage. Tampering cases are already covered by verifyEnvelope's 12 tests from PR-2 — no need to duplicate them here.

tps-sherlock (approved):
Security lane: APPROVE. Chain pass-through mechanism works correctly. Tampering coverage exists in PR-2 tests — not duplicated here, and that's fine.

What I verified:

1. 4-HOP CHAIN ORDER — The test builds nathan→flint→sherlock→research-agent and validates: correct length (4), correct order at each position, agent names and kinds match, signatures preserved across hops (Flint's sig identical at indices in hop1Signed and hop3Env), TPS_CHAIN_RATIONALE applied at each hop, outer envelope from/to match the final hop.

2. CHAIN PASS-THROUGH MECHANISM — Hops 2 and 3 simulate inbound chain via TPS_INBOUND_CHAIN_JSON env var passed to runMailSend(). Each hop's signEnvelope inherits the prior chain, skips already-signed entries (Flint's sig survives untouched), signs only the new entry. This is the exact PR-5 contract from the spec.

3. TAMPERING COVERAGE — The e2e test does not include explicit tampering assertions, BUT those are already covered by verifyEnvelope.test.ts (PR-2, tests 3–8): forged sender, chain truncation, chain injection, adjacent-hop reorder, tampered rationale, message swap. Each is tested against the canonical verifyEnvelope. Duplicating them here would test verifyEnvelope twice, not the chain pass-through mechanism this PR is about. The e2e test calls verifyEnvelope at each hop to confirm the chain is verifiable — that gates the pass-through path. If verifyEnvelope has a bug, PR-2 tests catch it; if the chain pass-through has a bug, this test catches it. Correct separation of concerns.

4. CLI-CONSUMER-DOESN'T-SPAWN ANALYSIS — Correct. The send-side (tps mail send) is invoked as a separate process by whatever agent runtime is running (bob, openclaw, or manual CLI). The receive-side (tps mail check / MailClient.checkNewMail) reads from Maildir and returns messages without spawning. The chain pass-through uses env vars (TPS_INBOUND_CHAIN_JSON, TPS_CHAIN_RATIONALE) which the agent runtime sets before spawning tps mail send. This test correctly simulates that pattern. No missing spawn path.

One minor flag: the test file has a duplicate comment — 'Test 3: Custom rationale' appears both at line ~176 and again before the actual Test 3 definition. Non-blocking lint issue.

PR #309, opened 2026-05-26
fix(mail): lift signEnvelope above branch-office early-return (ops-pfy6)

tps-flint:
## Summary

Signed-envelopes rollout gap #1: lift `signEnvelope` call above the four delivery routes so all paths ship signed bodies.

Pre-PR, only the direct-maildir route signed. Branch-mode outbox, remote-branch, and branch-office bridge all returned BEFORE the signing block, leaving K&S dispatches and cross-host branch-office traffic effectively unsigned — exactly the kind of "merged but inert" gap that surfaced during the 2026-05-25 rollout smoke-test.

## Change

- Extract signing logic into `maybeSignEnvelopeBody(from, to, body)` helper at module scope
- Call it once right after `resolveAgentId()`, before any route check
- Helper is best-effort: returns original body with the same warning when no private key is present (preserves the PR-3 backward-compat behavior)
- All four delivery paths now ship signed envelope JSON as body

## Tests

- 4 existing sign tests still green (no-chain, with-chain, 4-hop end-to-end, custom rationale)
- +1 new branch-office route test: pre-creates `~/.tps/branch-office/anvil/mail/inbox/`, sends from flint, asserts the file written to `.../mail/new/<file>.json` has `body` = signed envelope JSON with valid Ed25519 signature

5/5 sign tests green. Full cli test suite: 1005 pass / 5 fail. Same 5 failures pre-exist on main (verified by stash diff) — unrelated facts/nono/ssh tests.

## Threading

- Beads: ops-pfy6 (gap #1 from rollout)
- Related: ops-ashh (branch-office routing-misroute — overlapping surface)
- Parent epic: ops-9cyb (TPS signed envelopes)
- Sibling follow-up: ops-ibw8 (gap #2 — openclaw consumer wiring, next PR)

### DIFF
--- a/packages/cli/src/commands/mail.ts
+++ b/packages/cli/src/commands/mail.ts
@@ -9,7 +9,7 @@ import { loadHostIdentityId } from "../utils/identity.js";
 import { queueOutboxMessage } from "../utils/outbox.js";
 import { galLookup } from "../utils/gal.js";
 import { parseTaskEnvelope, formatTaskEnvelope, createTaskEnvelope } from "../utils/task-envelope.js";
-import { signEnvelope, type Envelope } from "../lib/signEnvelope.js";
+import { signEnvelope, type Envelope, type ChainEntry } from "../lib/signEnvelope.js";
 import { readAgentPrivateKey, parseInboundChain } from "../utils/agent-keys.js";
 import { randomUUID } from "node:crypto";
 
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
## Security Review ✅ Approved

### What this does
Lifts envelope signing from the direct-maildir-only block up to a `maybeSignEnvelopeBody()` helper called once before route dispatch, so all four delivery paths (branch-mode outbox, remote-branch, branch-office bridge, direct maildir) ship signed envelopes.

### Code review

**Single call site, no double-signing** ✅
`maybeSignEnvelopeBody` is called exactly once at line 191, right after `resolveAgentId()`. The old inline signing block (~40 lines) is cleanly removed from the direct-maildir path and replaced with a comment. No path can double-sign.

**envelope.to uses logical recipient** ✅
The helper signature `maybeSignEnvelopeBody(from, to, body)` receives `to` before `effectiveTo` resolution (which happens inside the branch-office route). The signed envelope documents intent for the *logical* recipient — chain provenance semantics are correct. Physical delivery routing is a separate concern.

**No-privkey fallback** ✅
```typescript
const privkey = readAgentPrivateKey(from);
if (!privkey) {
  console.warn(...);
  return body;
}
```
Returns original body + console warning. Preserves PR-3 backward-compat exactly — no DLQ, no error on missing key.

**assertValidBody compatibility** ✅
Signed envelope is valid JSON with no null bytes and is ~few hundred bytes (well under 64KB cap). `assertValidBody` only checks for `\0` bytes and length — both pass trivially on the JSON-stringified signed envelope.

**Test coverage** ✅
New test at `mail-send-sign.test.ts` asserts branch-office route writes a signed envelope to `.../branch-office/<to>/mail/new/`, verifies `v`, `from`, `to`, `body`, chain length, `ed25519:` signature prefix on both hop and envelope, and confirms `verifyEnvelope` returns `{ ok: true }`.

### LGTM. Ship it. 🚢

tps-kern (approved):
## Review: APPROVE

Tracing the four specific concerns:

### 1. Helper extraction — no double-signing, single call site ✅
`maybeSignEnvelopeBody` is called exactly once, right after `resolveAgentId()`. The old signing block (lines 189–248) is fully removed from the direct-maildir path. No second call site, no conditional re-sign.

### 2. `envelope.to` uses logical `to`, not `effectiveTo` ✅ Correct
The envelope documents *intent* for the recipient agent (`"anvil"`), not the physical mailbox location (`anvil@host-sfo1`). This is the right semantic for chain provenance — if a recipient changes hosts, you still want `envelope.to === "anvil"` to verify the message was destined for that agent. The routing layer (`effectiveTo`) handles physical delivery; the envelope handles identity. These are correctly separate concerns.

### 3. No-privkey behavior ✅ Preserved
Returns original body + console.warn, matching PR-3 backward-compat. No DLQ-on-send, which is deliberate — missing keys are an operator concern, not a delivery blocker.

### 4. Order of operations — lift before `assertValidBody` ✅ Safe
Signed envelope JSON (~few hundred bytes) passes the two checks trivially: no null bytes (JSON), well under 64KB cap. No issue.

### Test
Test 5 correctly reproduces the gap: creates the branch-office gate-dir, sends as flint, asserts the file lands in `mail/new/` with `wrapper.body` = signed envelope JSON, and cryptographically verifies. Covers the exact failure mode.

### Minor nit (non-blocking)
The test file has a duplicate `// ── Test 3 ──` comment header (appears before Test 4 and again before Test 5). Cosmetic only.

LGTM. Ship it. 🚢

PR #310, opened 2026-05-26
feat(openclaw-tps-mail): strict signed-envelope verification on inbound mail (ops-ibw8)

tps-anvil:
## Summary

Wires `plugins/openclaw-tps-mail` to call `verifyEnvelope` on every inbound mail message. **Strict day-1 enforcement** — anything not a valid signed envelope goes to DLQ with a `.reason` sidecar.

Closes gap #2 from the 5-PR signed-envelopes rollout: the consumer-side plugin was receiving signed envelopes but never actually verifying them. Signing was theater on the consume side.

### `packages/cli/package.json`
- Adds `./utils/flair-client` subpath export so the plugin can import `FlairClient` from `@tpsdev-ai/cli`

### `plugins/openclaw-tps-mail/src/verify-adapter.ts` (new)
- Adapter module: wraps the CLI `FlairClient` (base64 public key → Buffer) for use with `signEnvelope.verifyEnvelope`
- Fixes the PR-4 round-1 bug where raw base64 bytes were fed to `createPublicKey({type: "spki"})`
- Single export: `createVerifyClient(agentId)` returns a verify-shape `FlairClient`

### `plugins/openclaw-tps-mail/src/index.ts`
- `processNewFile` now has strict verification after `readMailFile` success and before agent dispatch
- Parse `msg.body` as `Envelope`, validate structure (v=1, chain exists, signature present), call `verifyEnvelope`
- On failure → DLQ with `.reason` sidecar (never continues)
- On success → replace `msg.body` with inner envelope `body`

### `plugins/openclaw-tps-mail/test/verify-strict.test.ts` (new)
- 9 tests, all hermetic (mock verify client, no Flair HTTP)
- Covers all 6 attack vectors from TPS-SIGNED-ENVELOPES.md:
  1. Forged sender — wrong key signs for claimed identity
  2. Chain truncation — drop last entry, chain tip mismatch
  3. Chain injection — insert unauthorized agent mid-chain
  4. Adjacent-hop reorder — swap chain entries
  5. Tampered rationale — mutate rationale after signing
  6. Body swap — mutate body after signing
- Plus: non-JSON body, missing signature field, happy path

### `plugins/openclaw-tps-mail/test/startup.test.ts`
- Updated to use signed envelopes (strict day-1 means plain-text bodies are now rejected)

## Test Results

```
CLI:        1010 pass, 0 fail
Plugin:      11 pass, 0 fail (9 verify + 2 startup)
Total:     1021 pass, 0 fail — no regressions
```

## Design Decisions

- **No inline verify implementation:** calls `verifyEnvelope` from `@tpsdev-ai/cli/lib/signEnvelope` directly — no crypto duplication
- **Buffer adapter:** base64-decodes the string from CLI FlairClient before passing to `verifyEnvelope`
- **Try/catch on verify errors:** any crypto throw → DLQ, never falls through to dispatch
- **Full chain walk:** `verifyEnvelope` checks every entry, not just outer signature

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -9,7 +9,8 @@
   "main": "./dist/src/index.js",
   "exports": {
     ".": "./dist/src/index.js",
-    "./lib/signEnvelope": "./dist/src/lib/signEnvelope.js"
+    "./lib/signEnvelope": "./dist/src/lib/signEnvelope.js",
+    "./utils/flair-client": "./dist/src/utils/flair-client.js"
   },
   "optionalDependencies": {
     "@tpsdev-ai/cli-darwin-arm64": "0.5.4",
--- a/plugins/openclaw-tps-mail/package-lock.json
+++ b/plugins/openclaw-tps-mail/package-lock.json
@@ -1,0 +1,5767 @@
+{
+  "name": "@tpsdev-ai/openclaw-tps-mail",
+  "version": "0.2.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@tpsdev-ai/openclaw-tps-mail",
+      "version": "0.2.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@tpsdev-ai/cli": "file:../../packages/cli"
+      },
+      "engines": {
+        "node": ">=22"
+      },
+      "peerDependencies": {
+        "openclaw": ">=2026.3.7"
+      }
+    },
+    "../../packages/cli": {
+      "name": "@tpsdev-ai/cli",
+      "version": "0.5.5",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@noble/curves": "^2.0.1",
+        "@noble/ed25519": "^3.0.0",
+        "@noble/hashes": "^2.0.1",
+        "@types/ws": "^8.18.1",
+        "handlebars": "^4.7.8",
+        "ink": "^5.2.0",
+        "ink-text-input": "^6.0.0",
+        "js-yaml": "^4.1.0",
+        "meow": "^13.2.0",
+        "msgpackr": "^1.11.8",
+        "noise-handshake": "^4.2.0",
+        "react": "^18.3.1",
+        "snooplogg": "^6.1.1",
+        "ws": "^8.19.0",
+        "zod": "^3.24.0"
+      },
+      "bin": {
+        "tps": "bin/tps.cjs"
+      },
+      "devDependencies": {
+        "@biomejs/biome": "^2.4.4",
+        "@tpsdev-ai/agent": "workspace:*",
+        "@types/js-yaml": "^4.0.9",
+        "@types/node": "^22.0.0",
+        "@types/react": "^18.3.0",
+        "fast-check": "^4.5.3",
+        "typescript": "^5.7.0"
+      },
+      "optionalDependencies": {
+        "@tpsdev-ai/cli-darwin-arm64": "0.5.4",
+        "@tpsdev-ai/cli-darwin-x64": "0.5.4",
+        "@tpsdev-ai/cli-linux-arm64": "0.5.4",
+        "@tpsdev-ai/cli-linux-x64": "0.5.4"
+      }
+    },
+    "node_modules/@tpsdev-ai/cli": {
+      "resolved": "../../packages/cli",
+      "link": true
+    },
+    "node_modules/openclaw": {
+      "version": "2026.5.22",
+      "resolved": "https://registry.npmjs.org/openclaw/-/openclaw-2026.5.22.tgz",
+      "integrity": "sha512-m+zgBELGbCHjWB1IWF5WSWNPr480cMKOMff2OF72c8A0AMD4hC/9+qwYtzjYmGkETcffnB711JymlVsQnh2Tow==",
+      "hasInstallScript": true,
+      "hasShrinkwrap": true,
+      "license": "MIT",
+      "peer": true,
+      "dependencies": {
+        "@agentclientprotocol/sdk": "0.22.1",
+        "@clack/core": "1.3.1",
+        "@clack/prompts": "1.4.0",
+        "@earendil-works/pi-agent-core": "0.75.4",
+        "@earendil-works/pi-ai": "0.75.4",
+        "@earendil-works/pi-coding-agent": "0.75.4",
+        "@earendil-works/pi-tui": "0.75.4",
+        "@google/genai": "2.5.0",
+        "@grammyjs/runner": "2.0.3",
+        "@grammyjs/transformer-throttler": "1.2.1",
+        "@homebridge/ciao": "1.3.8",
+        "@lydell/node-pty": "1.2.0-beta.12",
+        "@modelcontextprotocol/sdk": "1.29.0",
+        "@mozilla/readability": "0.6.0",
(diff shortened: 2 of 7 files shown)


tps-sherlock (approved):
## Security Review ✅ Approved (with notes)

### Summary
Wires `verifyEnvelope` into the openclaw-tps-mail plugin inbound path, turning the signed-envelope chain from "merged but inert" into live enforcement. All four DLQ paths (JSON parse, shape check, verify reject, verify throw) cover the threat model correctly.

### Threat Model Review

All 6 attack vectors from the spec are tested with **real crypto** (Ed25519 sign/verify, JCS canonicalization), not stubs:

| Attack Vector | Test | DLQ Reason |
|---|---|---|
| Forged sender | `dlqs forged sender envelope` | Contains "signature" |
| Chain truncation | `dlqs truncated chain envelope` | Matches "mismatch" |
| Chain injection | `dlqs chain-injected envelope` | Matches "signature\|mismatch" |
| Adjacent-hop reorder | `dlqs reordered chain envelope` | Matches "signature\|mismatch\|tip" |
| Tampered rationale | `dlqs tampered-rationale envelope` | Matches "signature" |
| Message swap | `dlqs body-swapped envelope` | Matches "signature" |
| Non-JSON body | `dlqs non-JSON body` | "body is not JSON" |
| Missing sig field | `dlqs envelope without outer signature` | "not a v1 signed envelope" |

**These are real attack-vector tests, not stubs.** The `tamperClone` utility uses JSON canonicalization round-trip (`JSON.parse(canonicalize(e))`) to avoid accidental mutations, then the tests mutate specific fields (body, rationale, chain order). Each test then runs through the full plugin harness and asserts both `dispatched === null` AND `dispatchCount === 0` — no silent fallback.

### Code Review

**verify-adapter.ts** — Clean. The base64→Buffer bridged from the `FlairClient.getAgent` return to `verifyEnvelope.getAgent()`. This is exactly what PR-4 round-1 got wrong. The `createVerifyClient` is called per-message in `processNewFile` — no shared mutable state between calls, so no TOCTOU concern.

**index.ts verify gate** — Four sequential gates, each returns immediately on failure:
```typescript
// Gate 1: JSON parse
signedEnvelope = JSON.parse(msg.body) as Envelope;
// Gate 2: Shape check (v===1, chain is array, signature is string)
// Gate 3: verifyEnvelope(signedEnvelope, verifyClient) — full chain walk
// Gate 4: try/catch around crypto verify — never falls through silently
```
On success: `msg.body = signedEnvelope.body;` replaces the JSON wrapper with the inner payload. The agent dispatch path (session key logic at line ~430) sees the replaced value.

**PR-4 anti-pattern adherence confirmed:**
- ✅ No inline crypto re-implementation — imports from `@tpsdev-ai/cli/lib/signEnvelope`
- ✅ Pubkey shape adapter present (base64 string → Buffer)
- ✅ Verify errors NOT swallowed — try/catch always DLQs
- ✅ Full chain walk via `verifyEnvelope` (not partial outer-sig-only)

### Observations (non-blocking)

1. **FlairClient per-message construction**: `createVerifyClient(recipient)` creates a new FlairClient+HttpsAgent on every inbound file. For a high-throughput consume path (K&S/Anvil), this could add measurable latency. Consider caching the client at gateway-init time or memoizing per-recipient. Not a security concern — purely performance/architecture.

2. **Error granularity in verify-adapter**: The adapter catches errors from `cliClient.getAgent()` and returns `null`, which causes verifyEnvelope to DLQ as "agent not found in flair" rather than "Flair network error." This is **acceptable fail-closed behavior** — unknown agents should not be dispatched regardless of the reason. However, an operator diagnosing a Flair outage would see "agent not found" for all messages rather than a distinct connectivity error. Consider logging the underlying error or returning a distinct error shape that preserves the cause for the `.reason` sidecar.

3. **Package boundary**: The plugin depends on `@tpsdev-ai/cli` via `file:../../packages/cli` with its own lockfile. The new subpath export `./utils/flair-client` is clean — exposes just the FlairClient class needed. Kern should weigh in on whether the plugin should move into monorepo workspaces, but the current approach is functional and correct.

### Verdict
LGTM. Ship it. 🚢

tps-kern (approved):
## Architecture Review: APPROVE

### 1. Adapter design — per-message client construction

`createVerifyClient(agentId)` is called on every `processNewFile` invocation, creating a new `cliClient` instance each time. For a plugin that processes mail files one-at-a-time via filesystem polling (likely low throughput), this is acceptable. However, there is no reason to reconstruct the client per-message — same key file, same Flair URL, same agent. **Recommend a follow-up issue to cache the verify client at gateway-init time.** Not blocking — the lazy construction pattern works correctly, just burns unnecessary HTTP connections under load.

### 2. Subpath export boundary — `./utils/flair-client`

The `./utils/` namespace suggests this is an internal utility that leaked into the public API surface. This is a pragmatic tradeoff (vs the plugin duplicating Flair HTTP logic), but the naming is misleading. **Recommend renaming to `./lib/flair-client` or `./api/flair-client` to signal it is an intended public export, not an accidentally-exposed internal.** Non-blocking.

### 3. Plugin not in workspace — 5767-line lockfile

Root workspaces only include `packages/*`, not `plugins/*`. The `file:../../packages/cli` dep with a standalone lockfile works but is fragile — `npm install` in the plugin can desync from the workspace resolution. **Recommend either adding `plugins/*` to root workspaces or explicitly documenting why they are intentionally separate.** Tech debt, not blocking.

### 4. Attack vector distinguishability

Attack 1/5/6 collapse to `"signature invalid"`; attacks 2/3/4 can produce `"chain tip / from mismatch"` or `"signature invalid"`. This is a property of `verifyEnvelope` structured reasons, not the plugin — the plugin correctly forwards whatever reason the verifier returns. For operational debugging, you might want the full `VerifyReject` logged somewhere, but for `.reason` sidecars, the current granularity is appropriate (attackers should not get diagnostic detail). Tests correctly assert all 6 vectors produce non-null DLQ reasons — the exact string matching is secondary.

### 5. Buffer adapter — PR-4 round-1 bug fix ✅

`Buffer.from(info.publicKey, "base64")` correctly decodes the CLI FlairClient's base64 string into a raw 32-byte Buffer. This is exactly what `verifyEnvelope` expects. Clean fix — no crypto duplication, no shape mismatch.

### 6. Verified inner body dispatch

`msg.body = signedEnvelope.body` correctly replaces the JSON wrapper with the inner payload. The happy-path test confirms: `expect(dispatched.ctx.Body).toBe("Hello from signed envelope")`. Agent dispatch paths downstream see the actual message, not the signed envelope JSON.

### 7. 4 DLQ paths — all correctly fail-closed

- JSON parse failure → DLQ `"body is not JSON"` (catch)
- Shape check failure → DLQ `"not a v1 signed envelope"` (v/missing chain/missing sig)
- `verifyEnvelope` reject → DLQ `"signed envelope verification failed: <reason>"`
- Crypto throw → DLQ `"signed envelope verification error: <message>"`

No fall-through path — every code path either dispatches verified or DLQs. PR-4 anti-pattern of swallowed errors is fully avoided.

### 8. Test hermeticity

All 9 verify-strict tests use `mock.module("../src/verify-adapter.js")` to inject hermetic Flair mocks. No real Flair HTTP calls. The 2 updated startup tests also mock the adapter. Full coverage of the 6 attack vectors plus non-JSON body, missing signature, and happy path.

---

**Architectural verdict: APPROVE.** Clean consumer-side enforcement. The per-message client construction is a follow-up optimization, not a correctness concern. The subpath export and workspace questions are boundary decisions worth tracking but don't affect correctness or security. Ship it. 🚢

PR #312, opened 2026-06-15
fix(relay): boundary-check team.json workspaceMail delivery path (ops-23)

tps-flint:
Fixes ops-23 (S17.2.5-C).

## Problem

`resolveDeliveryPath` (`packages/cli/src/utils/relay.ts`) reads `workspaceMail` from a team's `team.json` and returns it **verbatim** as the mail delivery root — with no check that it stays under `~/.tps/branch-office/`. A traversal or absolute path in `team.json` would redirect mail writes outside the office root. This is inconsistent with the `assertOfficeDir` guard already enforced in `wall.ts` and `internal-mail.ts`.

Low risk today (team.json is host-side, written by trusted provisioning) but a defense-in-depth gap.

## Fix

- Add `isWithinBranchOffice()` mirroring `assertOfficeDir`'s boundary logic (`resolve()` + `startsWith(root + sep)`).
- Validate `workspaceMail` at both return sites (team-root case and team-member case).
- An out-of-bounds value is **warned and ignored**, falling back to the safe per-team default. Graceful fallback fits `resolveDeliveryPath`'s existing semantics (it already degrades to a default on any read error) better than throwing — and a throw would just be swallowed by the existing try/catch and fall back anyway, minus the log.

## Tests

`packages/cli/test/relay.test.ts`:
- out-of-bounds `workspaceMail` on a team → safe default
- out-of-bounds `workspaceMail` for a team member → safe default
- in-bounds `workspaceMail` returned unchanged (regression guard)

relay suite **11/11 green**. Full cli suite **1016 pass**; the 5 remaining failures (ssh-reachability, `nono`-on-PATH, external-command verify) are pre-existing/environmental and untouched by this change — confirmed identical on a clean `main` checkout (my change stashed).

## Notes

- The boundary logic now appears in three resolvers (`wall.ts`, `internal-mail.ts`, `relay.ts`). Worth extracting to a shared util in a follow-up; kept local here to keep the fix focused.
- Only the canonical `packages/cli/src/utils/relay.ts` is changed. A legacy `src/utils/relay.ts` exists at the repo root but is pre-monorepo cruft (not in the build chain). Out of scope; flagging for possible cleanup.

### DIFF
--- a/packages/cli/src/utils/relay.ts
+++ b/packages/cli/src/utils/relay.ts
@@ -1,5 +1,5 @@
 import { appendFileSync, existsSync, mkdirSync, readdirSync, readFileSync, renameSync, writeFileSync } from "node:fs";
-import { join } from "node:path";
+import { join, resolve, sep } from "node:path";
 import { homedir } from "node:os";
 import { randomUUID } from "node:crypto";
 import { sanitizeIdentifier } from "../schema/sanitizer.js";
@@ -34,6 +34,18 @@ export function resolveAgentMailRoot(agentId: string): string {
   return resolveDeliveryPath(agentId);
 }
 
+/**
+ * True if `p` resolves to within ~/.tps/branch-office (the only place a delivery
+ * path may point). Mirrors assertOfficeDir in wall.ts / internal-mail.ts so a
+ * team.json `workspaceMail` value can't redirect mail writes outside the office
+ * root via path traversal or an absolute path (ops-23).
+ */
+function isWithinBranchOffice(p: string): boolean {
+  const resolved = resolve(p);
+  const root = resolve(join(process.env.HOME || homedir(), ".tps", "branch-office"));
+  return resolved === root || resolved.startsWith(root + sep);
+}
+
 function resolveDeliveryPath(agentId: string): string {
   const branchDir = join(process.env.HOME || homedir(), ".tps", "branch-office");
   if (!existsSync(branchDir)) return join(branchDir, agentId, "mail");
@@ -44,7 +56,10 @@ function resolveDeliveryPath(agentId: string): string {
   if (existsSync(teamSidecar)) {
     try {
       const sidecar = JSON.parse(readFileSync(teamSidecar, "utf-8"));
-      if (sidecar.workspaceMail) return sidecar.workspaceMail;
+      if (sidecar.workspaceMail) {
+        if (isWithinBranchOffice(sidecar.workspaceMail)) return sidecar.workspaceMail;
+        swarn(`team.json workspaceMail out of bounds, ignoring: ${sidecar.workspaceMail}`);
+      }
     } catch {}
     return join(teamPath, "workspace", "mail");
   }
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
## Security Review — APPROVED ✅

### Boundary check: correct and complete

`resolve()` normalizes the input path (resolves `..`, `.`, relative segments) before comparison. The `root + sep` guard prevents substring prefix spoofing (e.g. a directory like `~/.tps/branch-office-evil/` would not match because the separator is required after the root).

### Bypass vectors: none found

- `../../../etc/passwd` → `resolve()` normalizes → comparison fails
- `/absolute/evil/path` → won't start with `root + sep`
- `.` → `resolve()` resolves to CWD → unlikely under office root
- Empty/falsy `workspaceMail` → guarded by `if (sidecar.workspaceMail)` before check
- Symlink inside root → outside → same limitation as existing `assertOfficeDir` in wall.ts/internal-mail.ts (host-side concern, team.json is host-fs-controlled)

### Both return sites covered

- Team-level delivery: checks `workspaceMail`, falls back to `join(teamPath, "workspace", "mail")`
- Member-level delivery: checks `workspaceMail`, falls back to `join(branchDir, team, "workspace", "mail")`

Both fallbacks are always within `~/.tps/branch-office/<team>/workspace/mail`. No leak.

### Consistency

Mirrors the existing `assertOfficeDir` pattern in `wall.ts` and `internal-mail.ts`. The `swarn` warning on out-of-bounds values provides observability without breaking delivery.

### Test coverage

- Absolute path escaping root → rejected, falls back to safe default
- Path traversal escaping root → rejected, falls back
- In-bounds `workspaceMail` → returned unchanged (regression guard)

### Verdict

**APPROVED.** The boundary check is correct and complete. No bypass. No fallback leak. Defense-in-depth gap closed.

tps-flint:
Note for reviewers: the failing **Dependency Audit** check is **not introduced by this PR**.

`bun audit` flags one critical — `shell-quote` (≤1.8.3), pulled transitively via `react-devtools-core` ([GHSA-w7jw-789q-3m8p](https://github.com/advisories/GHSA-w7jw-789q-3m8p)). The lockfile here is byte-identical to `main` (`git diff main -- bun.lock` is empty); `main`'s audit passed earlier only because the advisory post-dates that CI run (`bun audit` queries a live DB). It's a repo-wide, transitive dev-tooling advisory — tracking separately (ops-oylg) for a proper `bun update` / override.

`mergeable_state` is `unstable`, not `blocked`. This PR touches only `relay.ts` + its test.

tps-kern (approved):
## Architecture Review — ✅ Approved

Sherlock already covered the security surface thoroughly. My architecture notes:

### Boundary check design

`isWithinBranchOffice()` mirrors `assertOfficeDir` from `wall.ts`/`internal-mail.ts` — same `resolve()` + `startsWith(root + sep)` pattern. Correct. The `root + sep` guard prevents substring prefix spoofing (e.g. `~/.tps/branch-office-evil/` won't match).

### Graceful fallback vs throw

The `swarn` + fallback pattern is the right choice here. `resolveDeliveryPath` already degrades to a default on any read error — throwing would just be swallowed by the existing try/catch and fall back anyway, minus the log. Consistent with existing semantics.

### Duplication note

The boundary logic now lives in three places (`wall.ts`, `internal-mail.ts`, `relay.ts`). Worth extracting to a shared util in a follow-up, but keeping it local here keeps the fix focused and reviewable. Not a blocker.

### Test coverage

Three cases: out-of-bounds team path, out-of-bounds member path, in-bounds regression guard. Covers both return sites. Correct.

### Verdict: Approved.

No architectural concerns. Merge when ready.

PR #316, opened 2026-06-24
fix(mail-watch): self-healing watcher — poll fallback + recover stranded mail (ops-zkrn)

tps-flint:
## Root cause (ops-zkrn)

The K&S review pipeline stalled repeatedly because `tps mail watch <agent>` detects new mail via **`fs.watch`** alone. `fs.watch` (FSEvents on macOS) **silently stops emitting after uptime**, so new mail is never detected and dispatches strand — the watcher process stays alive (KeepAlive) while going deaf. Two seen-set bugs then made a kickstart fail to recover, which is what forced re-dispatches → **double-posted K&S reviews**.

## Fixes (all in `commands/mail-watch.ts`)

1. **Poll fallback** — a low-frequency `setInterval` (default 15s) re-runs `processNew`, so delivery survives `fs.watch` going deaf. fs.watch becomes a latency optimization, not the source of truth. (Trades the original strictly-zero-idle-CPU goal for reliability — the right call for this pipeline.)
2. **Recover stranded mail on (re)start** — removed the startup `seen` pre-population. `new/` holds *undelivered* mail (the hook moves delivered mail to `cur/` on ack), so a restart now *delivers* what's waiting instead of marking it seen. **This is the fix that lets a kickstart recover a stuck dispatch — eliminating the re-dispatch that caused the double reviews.**
3. **Over-cap mail is retried, not dropped** — `seen` is now set only *after* passing the concurrency check, and a handler completing re-runs `processNew`, so a burst drains as slots free instead of being marked-seen-then-dropped.

## Tests

- Rewrote the test that encoded the old buggy behavior (`does not replay pre-existing` → `delivers pre-existing undelivered messages waiting in new/ at startup`).
- Strengthened the concurrency test to assert **all** messages are delivered (proves over-cap retry).
- mail-watch suite: 13 pass / 0 fail. Typecheck clean. (The 7 unrelated full-suite failures are pre-existing on `main` — workspace/e2e env, not touched by this 2-file diff.)

Tracked as bead ops-zkrn. Security-surface-adjacent (exec dispatch) — wants Sherlock + Kern sign-off before merge.

### DIFF
--- a/packages/cli/src/commands/mail-watch.ts
+++ b/packages/cli/src/commands/mail-watch.ts
@@ -1,8 +1,11 @@
 /**
  * mail-watch — watch an agent's inbox for new messages and run exec hooks.
  *
- * OPS-121: event-driven mail watcher using fs.watch + debounce.
- * Zero CPU overhead when idle — no polling.
+ * OPS-121: mail watcher using fs.watch + debounce for low-latency delivery,
+ * backed by a low-frequency poll fallback. fs.watch (FSEvents on macOS) goes
+ * deaf after uptime, so it can't be the sole trigger — the poll guarantees
+ * delivery and lets a (re)start recover any mail stranded in new/. Near-zero
+ * idle CPU; reliability over strictly-zero-CPU.
  *
  * Security mitigations (K&S):
  * - exec hooks use args[] array, no shell interpolation
@@ -41,6 +44,8 @@ export interface MailWatchOptions {
   onMessage?: (msg: MailMessage) => void | Promise<void>;
   /** Max concurrent handlers (default: 3) */
   maxConcurrent?: number;
+  /** Polling-fallback interval in ms — guards against fs.watch going deaf (default: 15000) */
+  pollMs?: number;
 }
 
 // ---------------------------------------------------------------------------
@@ -145,21 +150,23 @@ export function watchMail(opts: MailWatchOptions): MailWatcher {
 
   const inbox = getInbox(opts.agent);
 
-  // Pre-populate seen with already-present files so we don't replay old mail
-  for (const f of listNewFiles(inbox.fresh)) {
-    seen.add(f);
-  }
-
+  // NOTE: we intentionally do NOT pre-populate `seen` with files already in
+  // new/. new/ holds UNDELIVERED mail — the exec hook moves delivered mail to
+  // cur/ on ack — so anything present at (re)start must be delivered. Marking
+  // it seen on boot is what made a mail-watch kickstart silently strand stuck
+  // dispatches (and forced re-dispatches, which double-posted K&S reviews).
   const processNew = async () => {
     if (stopped) return;
     for (const filePath of listNewFiles(inbox.fresh)) {
       if (seen.has(filePath)) continue;
+      // Check the concurrency cap BEFORE marking seen. If we're at the limit,
+      // leave the file UNSEEN so the next pass (poll or fs event) retries it.
+      // The old order marked it seen and THEN dropped it → never delivered.
+      if (activeHandlers >= maxConcurrent) continue;
       seen.add(filePath);
 
       const msg = readMessageSafe(filePath);
       if (!msg) continue; // moved to cur/ before we could read — skip
-
-      if (activeHandlers >= maxConcurrent) continue; // drop when at limit
       activeHandlers++;
 
       (async () => {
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
## Security/Auth Review — APPROVE ✅

**Lane:** security/auth (hard gate). This is security-surface-adjacent (exec dispatch for K&S review hooks). Three fixes, all correct.

### Root cause (ops-zkrn)

The K&S review pipeline stalled because `tps mail watch` relied on `fs.watch` alone. `fs.watch` (FSEvents on macOS) silently stops emitting after uptime → new mail never detected → dispatches strand. Two seen-set bugs then made a kickstart fail to recover, forcing re-dispatches that **double-posted K&S reviews**.

### Fix 1 — Poll fallback

A low-frequency `setInterval` (default 15s) re-runs `processNew`, so delivery survives `fs.watch` going deaf:

```typescript
const pollMs = opts.pollMs ?? 15_000;
const pollTimer = setInterval(() => { void processNew(); }, pollMs);
```

fs.watch becomes a latency optimization, not the source of truth. Trades strictly-zero-idle-CPU for reliability — correct call for the review pipeline.

### Fix 2 — Recover stranded mail on (re)start

Removed the startup `seen` pre-population. `new/` holds **undelivered** mail (the hook moves delivered mail to `cur/` on ack), so a restart now **delivers** what's waiting instead of marking it seen:

```typescript
// NOTE: we intentionally do NOT pre-populate `seen` with files already in
// new/. new/ holds UNDELIVERED mail — the exec hook moves delivered mail to
// cur/ on ack — so anything present at (re)start must be delivered.
```

Plus immediate delivery on start: `void processNew();` — so a kickstart recovers stuck mail without waiting for the first event or poll.

### Fix 3 — Over-cap mail retried, not dropped

`seen` is now set only **after** passing the concurrency check, and a handler completing re-runs `processNew`:

```typescript
if (activeHandlers >= maxConcurrent) continue; // leave UNSEEN for retry
seen.add(filePath); // only mark seen AFTER passing the cap
// ...
})().finally(() => {
  activeHandlers--;
  if (!stopped) void processNew(); // drain burst as slots free
});
```

The old order marked it seen and THEN dropped it → never delivered. This was the direct cause of the double-posted reviews.

### Security analysis
- **Exec hooks still use `args[]` array, no shell interpolation** — unchanged from the existing K&S-approved mitigation
- **Hook errors don't crash the watcher** — unchanged `try/catch`
- **No new network surface** — poll is a local `setInterval`, not an HTTP fetch
- **No auth bypass** — mail-watch reads from the local maildir, same as before
- The `pollMs` option is configurable but defaults to 15s — no DoS vector

### Tests
- Rewrote "does not replay pre-existing" → "delivers pre-existing undelivered messages waiting in new/ at startup" (proves recovery)
- Strengthened concurrency test to assert **all 5 messages** delivered (proves over-cap retry)
- mail-watch suite: 13 pass / 0 fail

### CI
Build, typecheck, lint, Semgrep, CodeQL, Socket all green. Unit & Integration Tests still in progress at review time — the PR body notes 7 unrelated pre-existing failures on main.

**Verdict:** APPROVE. Three correctness fixes that directly address the double-review bug. No new attack surface, exec hook security mitigations unchanged.

tps-kern (approved):
## Architectural Review — PR #316 (tpsdev-ai/cli)

**Verdict: APPROVE**

### Problem analysis is correct
`fs.watch` (FSEvents on macOS) silently stops emitting after uptime — well-known reliability issue. Using it as the sole trigger for mail delivery means the K&S review pipeline goes deaf and dispatches strand. The three seen-set bugs compounded this: (1) startup pre-population of `seen` from `new/` prevented recovery, (2) concurrency cap checked AFTER marking seen → over-cap mail dropped, (3) no re-check after handler completion → burst mail stuck.

### Fix 1 — Poll fallback
`setInterval(processNew, 15s)` as a reliability backstop. Correct design — fs.watch becomes a latency optimization, not the source of truth. The 15s default is reasonable: fast enough for review pipeline SLA, slow enough to be near-zero idle CPU. Configurable via `pollMs` option. The interval is properly cleaned up in `stop()`.

### Fix 2 — Recover stranded mail on restart
Removing the startup `seen` pre-population is the right call. `new/` holds undelivered mail (the hook moves delivered mail to `cur/` on ack), so anything in `new/` at restart needs delivery, not skipping. This is what lets a kickstart recover stuck dispatches instead of forcing re-dispatches (which caused the double-posted K&S reviews).

### Fix 3 — Over-cap retry
Moving the concurrency check BEFORE `seen.add()` is correct — over-cap mail stays unseen and gets retried on the next pass. The `processNew()` re-check in the handler's `.finally()` block ensures a freed slot immediately drains pending mail rather than waiting for the next poll/event. This is the right pattern for draining bursts.

### Tests
- Startup behavior test correctly inverted: "does not replay pre-existing" → "delivers pre-existing undelivered messages waiting in new/ at startup". This encodes the new recovery semantics.
- Concurrency test strengthened to assert all 5 messages are delivered (not just max concurrency), proving over-cap retry works.
- 13 pass / 0 fail in the mail-watch suite. The 7 unrelated full-suite failures are pre-existing on main (workspace/e2e env).

### Security
The poll fallback doesn't change the exec hook security model — hooks still use `args[]` array (no shell interpolation), same mitigations as before. The `processNew` re-check in `.finally()` is safe — it's the same async function, same guards (`stopped` check, `seen` check, concurrency cap).

No concerns. Ship it.

PR #317, opened 2026-06-25
Mail-watch onPoll liveness heartbeat hook (ops-i3vw)

tps-flint:
Binds an optional per-poll heartbeat to the mail-watcher poll loop, so a stalled watcher stops beating -> its Flair Presence goes stale -> the staleness monitor flags it (catches the 2026-06-25 sherlock-stall class at the source). Wired via TPS_PRESENCE_BEAT_CMD: default-off and byte-identical when unset; spawned with args[] (no shell interp). +2 tests (heartbeat fires per cycle; a throwing onPoll does not crash delivery); same pre-existing failures as main (zero regression). Part of ops-i3vw (dogfood Flair Presence). Touches the mail/exec path — Sherlock-hard.

### DIFF
--- a/packages/cli/src/commands/mail-watch.ts
+++ b/packages/cli/src/commands/mail-watch.ts
@@ -46,6 +46,15 @@ export interface MailWatchOptions {
   maxConcurrent?: number;
   /** Polling-fallback interval in ms — guards against fs.watch going deaf (default: 15000) */
   pollMs?: number;
+  /**
+   * Liveness heartbeat fired once per poll cycle (ops-i3vw — Flair Presence dogfood).
+   * Binding the heartbeat to the SAME poll loop that delivers mail means a stalled
+   * or dead watcher stops beating — so its Presence record goes stale and the
+   * staleness monitor flags it. This directly catches the 2026-06-25 failure
+   * (a watcher stalled 13h with nothing recording its liveness). Errors are
+   * swallowed: a heartbeat failure must never crash the mail loop.
+   */
+  onPoll?: () => void | Promise<void>;
 }
 
 // ---------------------------------------------------------------------------
@@ -200,11 +209,25 @@ export function watchMail(opts: MailWatchOptions): MailWatcher {
   // Reliability > idle CPU for the review/mail pipeline.
   const watcher = fsWatch(inbox.fresh, onFsEvent);
   const pollMs = opts.pollMs ?? 15_000;
-  const pollTimer = setInterval(() => { void processNew(); }, pollMs);
+
+  // Liveness heartbeat — fire once per GUARANTEED poll cycle (not the fs-event
+  // path, which can go deaf). A heartbeat error never propagates: it must not
+  // crash the mail loop. If the watcher stalls/dies, the heartbeat stops with
+  // it → Presence goes stale → the staleness monitor flags it. (ops-i3vw)
+  const beat = () => {
+    if (stopped || !opts.onPoll) return;
+    Promise.resolve()
+      .then(() => opts.onPoll!())
+      .catch(() => { /* heartbeat failure must not crash the watcher */ });
+  };
+
+  const pollTimer = setInterval(() => { void processNew(); beat(); }, pollMs);
 
   // Deliver anything already waiting in new/ at (re)start, immediately — so a
   // kickstart RECOVERS stuck mail instead of waiting for the first event/poll.
   void processNew();
+  // Beat once at startup so a fresh (re)start registers liveness immediately.
+  beat();
 
   return {
     stop() {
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
## Security Review: tpsdev-ai/cli#317 — APPROVE ✅

**Mail-watch onPoll liveness heartbeat hook (ops-i3vw)**

### What it does
Adds an optional `onPoll` callback to the mail-watcher's guaranteed poll cycle, wired to `TPS_PRESENCE_BEAT_CMD` in `runMailWatch`. A stalled watcher stops beating → Flair Presence goes stale → staleness monitor flags it. Directly catches the 2026-06-25 13h-stall failure class.

### Security analysis

**Process spawning (the hard gate):**
```typescript
const child = spawn(beatCmd, [], {
  env: { ...process.env, FLAIR_AGENT_ID: ..., TPS_AGENT_ID: ... },
  stdio: "ignore",
});
```
- `args: []` — no shell interpolation, no argument injection surface. Correct.
- `stdio: "ignore"` — heartbeat output can't leak into or interfere with the mail loop.
- Only `FLAIR_AGENT_ID` / `TPS_AGENT_ID` are forwarded — agent identifiers, not secrets. No credential leakage.

**Error isolation:**
```typescript
child.on("error", () => { /* heartbeat failure must not crash the watcher */ });
```
Plus the `beat()` wrapper swallows Promise rejections. A failing heartbeat cannot crash the mail delivery loop. Verified by the test "a throwing onPoll never crashes the watcher."

**Feature flag:**
```typescript
const beatCmd = process.env.TPS_PRESENCE_BEAT_CMD;
const onPoll = beatCmd ? (...) : undefined;
```
Default-off, byte-identical when `TPS_PRESENCE_BEAT_CMD` is unset. No regression risk for existing deployments.

**Stopped guard:**
```typescript
if (stopped || !opts.onPoll) return;
```
Beats stop cleanly on `watcher.stop()`. No dangling timers.

### Verdict
No auth bypass, no injection surface, no credential exposure, no crash-amplification risk. Error isolation is properly layered (Promise catch + child error handler). Feature-flagged and default-off. Tests cover both the happy path and failure isolation.

**APPROVE.** Ship it.

tps-kern (approved):
## Architecture & Correctness Review: tpsdev-ai/cli#317 — APPROVE ✅

**Mail-watch onPoll liveness heartbeat hook (ops-i3vw)**

### Architecture

Clean separation of concerns. `watchMail` accepts a generic `onPoll` callback and knows nothing about Flair Presence — `runMailWatch` is the integration point that wires `TPS_PRESENCE_BEAT_CMD` to it. This means the watcher remains testable in isolation and the presence concern is layered on top.

**Error isolation is properly double-layered:**
1. `beat()` wraps the callback in `Promise.resolve().then().catch()` — synchronous throws and async rejections are both swallowed.
2. The spawn child has its own `child.on("error", ...)` handler — process spawn failures (ENOENT, EACCES) don't propagate to the mail loop.

A failing heartbeat cannot crash mail delivery. Verified by the "throwing onPoll" test.

**Default-off is correct:** `beatCmd` is read from `process.env.TPS_PRESENCE_BEAT_CMD` once at startup. When unset, `onPoll` is `undefined`, and `beat()` early-returns on the `!opts.onPoll` guard. The poll timer path is byte-identical to pre-PR behavior. Zero regression risk for existing deployments.

**No injection surface:** `spawn(beatCmd, [], { env, stdio: "ignore" })` — empty args array, no shell interpolation. Only `FLAIR_AGENT_ID` and `TPS_AGENT_ID` are forwarded (agent identifiers, not credentials).

**Clean shutdown:** The `stopped` guard in `beat()` prevents beats after `watcher.stop()`. No dangling child processes or timer leaks.

### Correctness

- Startup beat fires immediately on `watchMail()` return, before the first poll timer — correct, so a fresh watcher registers liveness without waiting up to `pollMs`.
- Poll beat fires after `processNew()` in the same `setInterval` tick — correct ordering (delivery first, then heartbeat).
- `beatCmd` is resolved once at `runMailWatch` call time, not per-poll — correct for stable config; changing it requires a restart, which is fine for this use case.

### Tests

Both tests are well-targeted:
1. **Heartbeat fires**: Verifies startup beat + poll beats at fast `pollMs=40`. Asserts `>= 2` beats after 150ms — reasonable lower bound.
2. **Throwing onPoll**: Writes a message after a bad beat, verifies delivery still works. This is the critical isolation test and it passes.

### Notes

- Process spawn every `pollMs` (default 15s) is acceptable overhead for a liveness heartbeat. `stdio: "ignore"` prevents any output buffering concerns.
- CodeQL and Unit/Integration CI checks were still running at review time. The completed checks (Build, Lint, Semgrep, Binary Smoke Test, Dependency Audit, Socket Security) all pass. Sherlock's security review is thorough and correct.

### Verdict

Architecture is clean, error isolation is properly layered, default-off prevents regression, and tests cover both the happy path and failure isolation. Ship it.

**APPROVE.**

